PR #2, opened 2020-06-23
CI tests to pytest

btseytlin:


maximlopin:
1. Shouldn't `test_force_vectors()` also be in the new tests?
https://github.com/mindsdb/mindsdb_native/blob/master/tests/ci_tests/tests.py#L26

2. And again, `assert x` can mean different things: `assert len(x) != 0`, `assert x != 0`, `assert x is not None`, `assert x is True`, and it's unknown which one of these you actually meant by doing `assert x`.
Good examples of explicit assertions are all of assert statements below and including this line:
https://github.com/mindsdb/mindsdb_native/pull/2/files#diff-27eb329a19e2e57da6f47204a201a255R371

George3d6:
So... what I don't understand is as follows:

* Why move the CI tests into unit tests ?

As in, these are more or less "integration" tests I'd say, they test the whole of mindsdb end-to-end. They won't be even related to the `Predictor` object once we move some/most of that stuff into `functional.py`. Granted, it's a semantic thing, but I'd like to keep them separate, I think. 

* Did you remove the parameter iteration ?

Previously the CI tests were iterating through a few options:

GPU (if exists) + CPU
Ludwig + Lightwood (though we removed the former temporarily due to time constraints)
... etc

Even if in principle travis only allows us to test 1 or two right now it might be worth keeping this logic and it can still be useful for testing stuff more thoroughly locally.

* How are the tests differentiated between windows and linux/osx ?

Previously that was differentiated by having the'fast_test' be winodws based and the 'full_test' aimed at linux ? (because windows doesn't support tensorflow that well, so no way of testing ludwig, windows is slower on travis so we can't test so many things... etc).

Other than that see @maximlopin's comments. From my end, I'm fine removing some assertions and just replacing them with nothing (aka asserting that the function doesn't crash and assuming the result returned is correct because validating "correct" is actually quite hard for some of these things, at least if you want the validation to be meaningful). But mindsdb never returns None/False, so just `assert returned_value` will probably contribute nothing.

btseytlin:
> 1. Shouldn't `test_force_vectors()` also be in the new tests?
>    https://github.com/mindsdb/mindsdb_native/blob/master/tests/ci_tests/tests.py#L26
> 2. And again, `assert x` can mean different things: `assert len(x) != 0`, `assert x != 0`, `assert x is not None`, `assert x is True`, and it's unknown which one of these you actually meant by doing `assert x`.
>    Good examples of explicit assertions are all of assert statements below and including this line:
>    https://github.com/mindsdb/mindsdb_native/pull/2/files#diff-27eb329a19e2e57da6f47204a201a255R371

`test_force_vectors` was never called in the old tests iirc

btseytlin:
> So... what I don't understand is as follows:
> 
> * Why move the CI tests into unit tests ?
> 
> As in, these are more or less "integration" tests I'd say, they test the whole of mindsdb end-to-end. They won't be even related to the `Predictor` object once we move some/most of that stuff into `functional.py`. Granted, it's a semantic thing, but I'd like to keep them separate, I think.
> 
> * Did you remove the parameter iteration ?
> 
> Previously the CI tests were iterating through a few options:
> 
> GPU (if exists) + CPU
> Ludwig + Lightwood (though we removed the former temporarily due to time constraints)
> ... etc
> 
> Even if in principle travis only allows us to test 1 or two right now it might be worth keeping this logic and it can still be useful for testing stuff more thoroughly locally.
> 
> * How are the tests differentiated between windows and linux/osx ?
> 
> Previously that was differentiated by having the'fast_test' be winodws based and the 'full_test' aimed at linux ? (because windows doesn't support tensorflow that well, so no way of testing ludwig, windows is slower on travis so we can't test so many things... etc).
> 
> Other than that see @maximlopin's comments. From my end, I'm fine removing some assertions and just replacing them with nothing (aka asserting that the function doesn't crash and assuming the result returned is correct because validating "correct" is actually quite hard for some of these things, at least if you want the validation to be meaningful). But mindsdb never returns None/False, so just `assert returned_value` will probably contribute nothing.


Just moving them so that all tests are called by the same command so that we have less lines in CI code and for general cleaniness. These tests are not any different from other slow unit tests that can be found in `test_predictor.py` so I don't see a point in keeping them separate. 

I did not remove the parameter iteration. The test runs with `use_gpu=True` and `use_gpu=False` (see the `mark.parametrize`) line. The backend iteration I did not include because the old tests didn't run for any backend except lightwood.

Honestly I found no difference between `fast_test` and `full_test`. The "full test" used to run the tests in  `integration_tests` folder, but these were moved to pytest and after that there was no difference. 

Mindsdb doesn't return None/False so `assert something` is a good way to check that it in fact doesn't. I agree with Maxim that asserting something more detailed is always better though

btseytlin:
> 1. Shouldn't `test_force_vectors()` also be in the new tests?
>    https://github.com/mindsdb/mindsdb_native/blob/master/tests/ci_tests/tests.py#L26
> 2. And again, `assert x` can mean different things: `assert len(x) != 0`, `assert x != 0`, `assert x is not None`, `assert x is True`, and it's unknown which one of these you actually meant by doing `assert x`.
>    Good examples of explicit assertions are all of assert statements below and including this line:
>    https://github.com/mindsdb/mindsdb_native/pull/2/files#diff-27eb329a19e2e57da6f47204a201a255R371

Actually there are times when you just want to check all of the things mentioned in your point 2. That is "this thing didn't return something invalid". In that case `assert x` is great.

PR #3, opened 2020-06-24
Refactor predictor, move static methods to libs.controllers.functional

### LINKED ISSUE #437: Add static interface for method that don't need a predictor
Title says it all, not urgent.

btseytlin:
Resolves: https://github.com/mindsdb/mindsdb/issues/437

* Add a decorator for deprecated methods/classes/functions
* Move various methods of `Predictor` to functions `controllers.functional` 
* The `Predictor` methods are kept for backwards compatibility, we should probably remove them in a major release
* Stuff refactored


mindsdb/libs/controllers/functional.py
```diff
@@ -0,0 +1,178 @@
...
+def rename_model(old_model_name, new_model_name):
+    """
+    If you want to rename an exported model.
+
+    :param old_model_name: this is the name of the model you wish to rename
+    :param new_model_name: this is the new name of the model
+    :return: bool (True/False) True if predictor was renamed successfully
+    """
+
+    if old_model_name == new_model_name:
```
maximlopin: I think we should raise an exception when you attempt to rename a model that doesn't exist

maximlopin: it can be done after merge I guess, everything else looks good to me.

George3d6:
So, few notes, the import path should be one of two options:

`mindsdb.functional` and `mindsdb.F` (to match the pytroch interface, might as well use that as a guideline)

see the `__init__.py` in mindsdb for how to do that

Also, maybe worth modifying the current tests to check these changes (importing from mindsdb.functional).

Also, are you sure the name dict usage won't affect anything ? seems not to, since all the tests are passing.

other than that, looks perfect, thanks :)

mindsdb/libs/controllers/functional.py
btseytlin: Done

George3d6 (approved):

PR #19, opened 2020-06-25
Dependency graph

maximlopin:
1. I think logic is correct, but it takes forever to run.
2. The first run where input package is mindsdb will result in downloading 6GB~ of packages.
3. Packages don't provide their metadata (requirements & license) consistently (e.g. some .whl packages have metadata.json, some not)


George3d6:
@torrmal try running this when you have time.

PR #24, opened 2020-06-29
New sampling interface

### LINKED ISSUE #4: Automatic sampling & transform cache disabling
We need to:

a) Start sampling training data when the dataset is too large (e.g. can barely fit in RAM)
b) Automatically disable the transformed cache for lightwood when the dataset is too large (e.g. ~5-10% or more of the total RAM)

For now let's say we have 4 potential state for the data:
* small
* medium
* big
* huge

Determined as a how much of the available RAM the pandas dataframe uses.

If small --- do nothing
If medium --- use the current sampling algorithm to sample the data analysis data
If big --- same as medium + set transform cache disable in the lightwood configuration for training
If huge --- same as big + use the current sampling algorithm to sample data for the testing, training and validation sets.

### LINKED ISSUE #5: New sampling interface
We should have a new sampling interface that's a dictionary (named dictionary ?) that can be passed to `.learn`, it should look something like:
```
sample_settings = {
'sample_for_analysis`: None
,`sample_for_training`: None
,'sample_margin_of_error': 0.01
,'sample_percentage': None
,'sample_function': None
}
```

See #4 for more details on how these arguments will behave. The gist of it is:

None == default behavior for both analysis and training sampling as specified in #4 

* If `sample_function` is specified (should be a function pointer) that wil be used for sampling from the dataset for both the data analysis and the training sample (if they are enabled from their respective arguments for if we automatically detect they should be enabled based on #4)

* Else If `sample_percentage` is specified that will be used to randomly sampling from the dataset for both the data analysis and the training sample (if they are enabled or if we automatically detect they should be enabled)

* Else If `sample_margin_of_error` is specified that will be used for our internal sampling algorithm for both the data analysis and training sampling (if they are enabled or if we automatically detect they should be enabled)

* Else use the default `sample_margin_of_error` of `0.01` + the behavior specified above

btseytlin:
Resolves https://github.com/mindsdb/mindsdb_native/issues/5 and https://github.com/mindsdb/mindsdb_native/issues/4

When `Predictor.learn` or `Predictor.analyze_dataset` are called, the size of input data frame and total available ram is used to set the values of `sample_for_analysis`, `sample_for_training` and disabling lightwood transform cache. We can control at which point each of these optimization is enabled in this code:
https://github.com/mindsdb/mindsdb_native/pull/24/files#diff-bba6c06f10b48c23d6d7319bd1f48f69R23

User can override any of these by providing them as a `sampling_settings` parameter.

`sample_function` is stored in `hmd`


btseytlin:
Also discovered that tests were not really isolated because `MINDSDB_STORAGE` was the same across all tests. Fixed it, now tests use a temporary dir for storage.

Also helps by not leaving test models in your mindsdb storage dir, which is good.

tests/integration_tests/data_sources/test_maria_ds.py
```diff
@@ -77,8 +77,8 @@ def test_maria_ds():
     def assert_expected_type(column_typing, expected_type, expected_subtype):
         assert column_typing['data_type'] == expected_type
         assert column_typing['data_subtype'] == expected_subtype
-        assert column_typing['data_type_dist'][expected_type] == 199
-        assert column_typing['data_subtype_dist'][expected_subtype] == 199
+        assert column_typing['data_type_dist'][expected_type] == 200
```
maximlopin, line 82: creating a constant variable and referring to it everywhere else would remove confusion about where these numbers come from (`== 199`, `== 200`)

maximlopin: everything else looks good

Issue #12, opened 2020-06-25
Only use stats_v2 internally


George3d6:
Only use `stats_v2` internally instead of using the old `column_stats`.

PR #26, opened 2020-06-30
Remove column_stats and scores

btseytlin:
Resolves https://github.com/mindsdb/mindsdb_native/issues/12

:fire: I removed all scores! I don't know if any of these changes break some interfaces that frontend relies on. 

@George3d6 please review that carefully


George3d6:
No obvious flaws and the tests pass... I assume changes might be required to `mindsdb` itself to handle the new schema but I can make those, so no worries on that front. Merging.

Issue #29, opened 2020-07-01
Make sure lock is explicitly removed before closing


George3d6:
If there is a crash, in some situations, the lock file might remain open and the lock won't be removed.

We can fix this in two ways:

a) Add a try/except/finally to everything in functional.py and everything in controllers.py that uses a lock.

b) Add an atexit to everything mentioned above

Option `a` is annoying because we'd loss the crash logs and would have to add custom code (using traceback) to recover them.

Option `b` seems fairly good actually, but I don't know, maybe I'm missing something.

btseytlin:
> we'd loss the crash logs and would have to add custom code (using traceback) to recover them.

It's not true, just reraise the exception and no logs are lost

George3d6:
@btseytlin are you sure ? I believe we need the traceback as per @maximlopin current implementation.

If you are sure this is not the case and we find the time we should discuss it in order to simplify the code, for now I will close this issue though, as @maximlopin implementation is enough for our current needs.

PR #30, opened 2020-07-01
Locking decorator

maximlopin:
https://github.com/mindsdb/mindsdb_native/issues/29


mindsdb_native/libs/controllers/predictor.py
```diff
@@ -87,22 +82,23 @@ def prepare_sample_settings(self,
...
               ignore_columns=None,
               stop_training_in_x_seconds=None,
               backend='lightwood',
               rebuild_model=True,
               use_gpu=None,
               equal_accuracy_for_all_output_categories=True,
               output_categories_importance_dictionary=None,
-              unstable_parameters_dict=None,
-              sample_settings=None):
+              unstable_parameters_dict=None):
```
btseytlin: Careful, something is wrong. You are removing `sample_settings` and bringing back `sample_margin_of_error`

mindsdb_native/libs/controllers/predictor.py
```diff
@@ -162,12 +153,7 @@ def learn(self,
         test_from_ds = None if test_from_data is None else getDS(test_from_data)
 
         transaction_type = TRANSACTION_LEARN
-
-        sample_for_analysis, sample_for_training, disable_lightwood_transform_cache = get_memory_optimizations(
-            from_ds.df)
-        sample_settings, sample_function = self.prepare_sample_settings(sample_settings,
-                                                    sample_for_analysis,
-                                                    sample_for_training)
+        sample_confidence_level = 1 - sample_margin_of_error
```
btseytlin: Same here, you are overwriting the changes of my PR

mindsdb_native/libs/helpers/locking.py
```diff
@@ -1,34 +1,81 @@
...
+                final_lock_name = '{}_{}.lock'.format(lock_name, argval)
+
+            path = os.path.join(CONFIG.MINDSDB_STORAGE_PATH, final_lock_name)
+            f = open(path, 'a+')
+            portalocker.lock(f, flags)
+
+            try:
+                ret = func(*args, **kwargs)
+            except BaseException:
+                portalocker.unlock(f)
```
btseytlin: Unlocking can go into `finally` clause, and then the whole `except` clause can be omitted

maximlopin: I'm re-raising the exception under `except:`, will `finally` execute before it?

btseytlin: I think finally executes after everything else (even if you do `return` in any clauses). But I think if you move unlocking to finally, then you don't need the `except` clause at all as you don't need to raise

maximlopin: just checked that, `finally` indeed executes before `except:`

maximlopin:
OSX fails with `The command pip3 install --no-cache-dir -e . exited with 1.`

mindsdb_native/libs/helpers/locking.py
```diff
@@ -1,34 +1,77 @@
...
+            else:
+                return ret
+            finally:
+                portalocker.unlock(f)
+
+        return wrapper2
+    return wrapper1
+
+
+# def learn_lock(name):
```
btseytlin: Comments

btseytlin:
How about some tests?

I am mainly concerned about proper lock releasing. I think it should be fairly easy to test that no locks are in place after something is completed. 

I think it should be enough to test the decorator on a dummy function

mindsdb_native/libs/helpers/locking.py
```diff
@@ -1,34 +1,77 @@
...
+                if argname.startswith('self.'):
+                    argval = getattr(args[0], argname.split('.')[1])
+                elif argname in kwargs:
+                    argval = kwargs[argname]
+                else:
+                    index = inspect.getfullargspec(func).args.index(argname)
+                    if index < len(args):
+                        argval = args[index]
+                    else:
+                        raise Exception('argname wasn\'t found in *args/**kwargs')
```
George3d6: I fail to see why we need this bit of logic. Why not make `argname` mandatory and just name the lock `{lock_name}_{argname}.lock`.

After all `argname` basically ought to be the name of the Predictor we are locking, that should be a name we always have access to (maybe even rename it to this)

George3d6: ahhh... nvm, I just realized, this is because the decorator is external to the scope :(

Well, for once, I think you should removed the None check and not allow it to be none, but second, this makes things a bit complicated.

Are you sure the logic for getting `argname` here always works ? If not, would it be hard to replace with decorator with a `with` in order to simplify the logic ?

mindsdb_native/libs/helpers/locking.py
```diff
@@ -1,34 +1,77 @@
...
+                final_lock_name = '{}_{}.lock'.format(lock_name, argval)
+
+            path = os.path.join(CONFIG.MINDSDB_STORAGE_PATH, final_lock_name)
+            f = open(path, 'a+')
+            portalocker.lock(f, flags)
+
+            try:
+                ret = func(*args, **kwargs)
+            except BaseException:
```
George3d6: Why except "BaseException" here only ?

I think this should just be `except` since you want to catch everything, intrerrupts included, so that you can unlock the lock at exit, no matter the cause of existing.

mindsdb_native/libs/helpers/locking.py
```diff
@@ -1,34 +1,77 @@
...
-
-def get_data_lock(name):
-    delete_lock_fn = os.path.join(CONFIG.MINDSDB_STORAGE_PATH,f'{name}_delete.lock')
-    f = open(delete_lock_fn, 'a+')
-    portalocker.lock(f, portalocker.LOCK_SH+portalocker.LOCK_NB)
-    return f
-
-def unlock(file):
-    portalocker.unlock(file)
+def mdb_lock(flags, lock_name, argname=None):
```
George3d6: Can you make the flag one of the two options (strings) `shared` and `exclusive`

Then add up the portalocker flags in here (SH + NB for shared, EX + NB for exclusive)

We always want non blocking locks so there's no need to specify that.

George3d6 (changes requested):

mindsdb_native/libs/helpers/locking.py
btseytlin: `finally` clause exists for this case
https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions

maximlopin: 1. `BaseException` catches all possible exceptions (including `KeyboardInterrupt`), I'm using it over `except:` because it's more explicit.

2. `finally:` executes before `except:`, so the lock is unlocked even if there is a new exception raised under `except:`

mindsdb_native/libs/helpers/locking.py
maximlopin: "Well, for once, I think you should removed the `None` check and not allow it to be none, but second, this makes things a bit complicated."
Allowing it to be `None` might be useful if we for example want to lock `Predictor.learn` no matter what Predictor.name is equal to.

If we use `with` statement, then we have to do it everywhere. The goal of this decorator is to be able to just put it over any method or function, and whole locking logic is abstracted away.

Getting function argument from withing the decorator by the name of the argument is kind of tricky, I'm sure it's possible to do (with `inspect` module) but I'm not 100% certain that this implementation will not ever fail. We need some tests for it.

mindsdb_native/libs/helpers/locking.py
maximlopin: done

maximlopin:
working on a test for this

mindsdb_native/libs/helpers/locking.py
btseytlin: See link above: finally executes after all `except` clauses:
> If a finally clause is present, the finally clause will execute as the last task before the try statement completes.

This is the exact use-case of finally

mindsdb_native/libs/helpers/locking.py
George3d6: > Allowing it to be None might be useful if we for example want to lock Predictor.learn no matter what Predictor.name is equal to.

We will never have to do this, but we might by mistake, hence why I'm saying None should be removed

> If we use with statement, then we have to do it everywhere. The goal of this decorator is to be able to just put it over any method or function, and whole locking logic is abstracted away.

Fair enough, seem like a preference between indentation vs having to use function declarations inside a function declaration ,whichever works, I'm fine sticking with the decorator.

So, if you can get the tests rolling, prefer. You only need to test with the function names for which it is used.

George3d6: Ah, nevermind, I see you change to a `with` now :P

tests/unit_tests/libs/helpers/locking.py
```diff
@@ -0,0 +1,22 @@
...
+def test_locking():
+    for flags in ['shared', 'exclusive']:
+        lock = MDBLock(flags, 'name')
+
+        with pytest.raises(ExceptionForTest):
+            lock(func)()
+
+        with pytest.raises(ExceptionForTest):
+            with lock:
+                func()
```
btseytlin: I don't understand: does this assert something? Does this test validate that the lock is released properly?

George3d6: I *assume* it asserts (in the `exclusive` case) that the lock has been properly released previously.

But I agree that this test doesn't seem to do anything in the `shared` case.

@maximlopin maybe add some comments as to what the test is doing, or just remove the `shared` case and we can merge the PR without it.

*Also, might be cleaner to define `func` within `test_locking`.

George3d6:
Other than that, tests are passing, code looks fine, @maximlopin once you address that tiny issue with the test please go ahead and merge.

PR #31, opened 2020-07-01
Cleanup for tests, transaction output types and explanation

George3d6:
* Made the members of `PredictTransactionOutputData` and `TransactionOutputRow` that we wouldn't necessarily want to expose to the user "private"
* No longer testing for the existence of these members as part of testing the interface
* Remove redundant information from the `.explain()` function (as a result it will also be removed from `explanation`)
* Removed redundant checks and checks without explicit assertions from the `.explanation` tests and changed to checks for existence + python type correctness
* Renamed `epitomize` to `summarize` and made it print the confidence range instead of the predicted values for numbers

### DIFF
--- a/tests/unit_tests/libs/controllers/test_predictor.py
+++ b/tests/unit_tests/libs/controllers/test_predictor.py
@@ -106,8 +106,8 @@ class TestPredictor:
         result = mdb.predict(when={"numeric_x": 10, 'categorical_x': 1})
         explanation_new = result[0].explanation['numeric_y']
         assert isinstance(explanation_new['predicted_value'], int)
-        assert explanation_new['confidence_interval'][0] < explanation_new['predicted_value'] < explanation_new['confidence_interval'][1]
-        assert explanation_new['confidence'] >= 0.8
+        assert isinstance(explanation_new['confidence_interval'],list)
+        assert isinstance(explanation_new['confidence_interval'][0],int)
         assert isinstance(explanation_new['important_missing_information'], list)
         assert isinstance(explanation_new['prediction_quality'], str)
 
--- a/tests/unit_tests/libs/controllers/test_predictor.py
+++ b/tests/unit_tests/libs/controllers/test_predictor.py
@@ -100,7 +100,7 @@ class TestPredictor:
         mdb.learn(
             from_data=input_dataframe,
             to_predict='numeric_y',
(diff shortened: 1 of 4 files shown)


George3d6:
Note: The OSX test seems to be failing but for reasons unrelated to this, it also fails on staging, so I think we can consider the test are passing. I've tried looking into why, it seems like it's just a travis related issue where the install times out.

maximlopin (approved):
all looks good

tests/unit_tests/libs/controllers/test_predictor.py
```diff
@@ -334,37 +336,21 @@ def test_house_pricing(self, use_gpu):
...
 
-        def assert_prediction_interface(prediction):
-            assert hasattr(prediction, 'data')
-            assert hasattr(prediction, 'extra_insights')
-            assert hasattr(prediction, 'transaction')
-            assert hasattr(prediction[0], 'explanation')
-            assert hasattr(prediction[0], 'data')
-            assert prediction[0].predict_columns[0] == 'rental_price'
-
-            for item in prediction:
```
btseytlin, line 345: Why remove this part? I see that `__getitem__` interface is still there, so why not test it?

tests/unit_tests/libs/controllers/test_predictor.py
```diff
@@ -100,16 +100,18 @@ def test_explain_prediction(self):
         mdb.learn(
             from_data=input_dataframe,
             to_predict='numeric_y',
-            stop_training_in_x_seconds=1
+            stop_training_in_x_seconds=15
```
btseytlin: Is it required to train for so long? I think this can be as short as possible if the test only tests the interfaces

George3d6: No, it doesn't, I don't remember why I increased it, will reduce this number.

tests/unit_tests/libs/controllers/test_predictor.py
George3d6: I've made the members private (i.e. from `member` to `_member`) to indicated that they shouldn't be tested.

There's not point in testing for the existence of private member... even testing for public members is overkill, but I can see the point there, since a user may wish to access them and maybe someone took them out by mistake.

PR #34, opened 2020-07-02
Fix OSX tests

George3d6:
* Upgraded to xcode12

* Installing everything as `--user`

* Consequently, running `pytest` with `python3 -m` since it's only installed for the user

* Added short sleep after the mindsdb_native installation for OSX

Honestly, it might be that one or multiple of these changes are *not* required, I'm only ~100% certain the `--user` one is, but testing with travis is slow and hard and prone to error because traivs sometimes "naturally" fails or delays builds by 5-10 minutes.

So, unless anyone wants to have fun experimenting, this should be merged as is for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,9 +54,9 @@ install:
   - if [ "$TRAVIS_OS_NAME" = "windows" ]; then travis_wait 15 pip install --no-cache-dir -e .; fi
   - if [ "$TRAVIS_OS_NAME" = "windows" ]; then travis_wait 15 pip install --no-cache-dir -r requirements_test.txt; fi
   - if [ "$TRAVIS_OS_NAME" = "windows" ]; then travis_wait 15 pip install --no-cache-dir -r optional_requirements_extra_data_sources.txt; fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -54,16 +54,21 @@ install:
...
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then travis_wait 15 pip3 install --no-cache-dir -r optional_requirements_extra_data_sources.txt; fi
+
 # run tests
 before_script:
   # Travis doesn't support docker on OSX and it doesn't work on Windows too
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then travis_wait 20  docker-compose up -d; fi
 script:
   # Unit tests
-  - travis_wait 15 pytest --run-slow --durations=20 tests/unit_tests
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then travis_wait 15 python3 - m pytest --run-slow --durations=20 tests/unit_tests; fi
```
btseytlin, line 70: I am sure this is not required

George3d6: Se my description, `pyest` needs to be installed as user on OSX so you don't get the `pytest`  comamnd, you have to explicitly call it as a module.

PR #35, opened 2020-07-02
Remove unneeded random seed setting is tests

btseytlin:
It's already fixed by pytest-randomly plugin, so we don't need to fix it in code


btseytlin:
Also noticed some fixture in tests that does nothing, my bad

PR #37, opened 2020-07-03
move text analysis from type_deductor to data_analyzer

maximlopin:
text analysis is now in `DataAnalyzer` and stored in stats_v2

### DIFF
--- a/mindsdb_native/libs/helpers/text_helpers.py
+++ b/mindsdb_native/libs/helpers/text_helpers.py
@@ -16,28 +16,26 @@ import json
 import hashlib
 import numpy
 import flair
+import langdetect
+langdetect.DetectorFactory.seed = 0
 
 
 def get_language_dist(data):
     lang_dist = defaultdict(lambda: 0)
     lang_probs_cache = dict()
-    try:
-        # @TODO There's repeat code here, is transformation to `flair.data.Sentence` quick enough that we don't care ?
-        for text, sent in zip(data, map(flair.data.Sentence, data)):
-            if text not in lang_probs_cache:
-                try:
-                    lang_probs = langdetect.detect_langs(text)
-                except langdetect.lang_detect_exception.LangDetectException:
-                    lang_probs = []
-                lang_probs_cache[text] = lang_probs
+    for text in data:
+        if text not in lang_probs_cache:
+            try:
+                lang_probs = langdetect.detect_langs(text)
+            except langdetect.lang_detect_exception.LangDetectException:
+                lang_probs = []
+            lang_probs_cache[text] = lang_probs
 
-            lang_probs = lang_probs_cache[text]
-            if len(lang_probs) > 0 and lang_probs[0].prob > 0.90:
-                lang_dist[lang_probs[0].lang] += 1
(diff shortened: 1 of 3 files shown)


mindsdb_native/libs/phases/data_analyzer/data_analyzer.py
```diff
@@ -281,6 +283,14 @@ def run(self, input_data):
                 stats_v2[col_name]['nr_warnings'] += 1
             self.log.info(f'Finished analyzing column: {col_name} !\n')
```
George3d6: This should come after the final `if`

George3d6: Actually, the whole of adding the text analysis to stats-v2 should come before the above loop that logs the warnings.

mindsdb_native/libs/helpers/text_helpers.py
```diff
@@ -10,9 +10,38 @@
...
+import flair
+
+
+def analyze_sentences(data):
+    """
+    :param data: list of str
+    :returns: 
+    tuple(
+        int: nr words total,
+        tuple(
```
George3d6: this is not a tuple (anymore), it's just 3 different parameters, feel free not to leave comments for everything, is a function is small we can read it, incorrect comments do more harm than no comments.

mindsdb_native/libs/helpers/text_helpers.py
```diff
@@ -10,9 +10,38 @@
...
+            dict: word_dist
+            dict: nr_words_dist
+        )
+    )
+    """
+    nr_words = 0
+    word_dist = Counter()
```
George3d6: Why make this a `Counter ?

mindsdb_native/libs/helpers/text_helpers.py
```diff
@@ -10,9 +10,38 @@
...
+    nr_words_dist = Counter()
+    for text in data:
+        sent = flair.data.Sentence(str(text))
+        nr_words_dist[len(sent)] += 1
+        nr_words += len(sent)
+        for tok in sent:
+            word = tok.text.strip(string.punctuation + '"\'«»')
+            word_dist[word] += 1
+
+    return nr_words, word_dist, nr_words_dist
```
George3d6: [Optional]

Can probably skip adding to `nr_words` and instead just do something like:

`nr_words = sum([k*v for k,v in nr_words_dist.items()])`

George3d6 (changes requested):

mindsdb_native/libs/helpers/text_helpers.py
maximlopin: should be `defaultdict(lambda: 0),` fixed

maximlopin:
`tests\unit_tests\libs\phases\type_deductor\test_type_deductor.py` fails:
```
assert stats_v2[col_name]['typing']['data_type'] == expected_type
E           AssertionError: assert 'Categorical' == 'Text'
```

_update:
fixed, local tests passed_

btseytlin (approved):

PR #40, opened 2020-07-06
test text analysis

maximlopin:
test logic of word dist, sentence length dist, language dist

### DIFF
--- a/tests/unit_tests/libs/helpers/text_helpers.py
+++ b/tests/unit_tests/libs/helpers/text_helpers.py
@@ -15,15 +15,15 @@ def test_language_analysis():
         'de': ['führen', 'Stelle', 'heißen', 'konnten', 'schlimm', 'mögen', 'Nähe'],
     }
 
-    m = 7
-    n = 10
+    sent_size = 7
+    num_sents = 10
 
     for lang, words in WORDS.items():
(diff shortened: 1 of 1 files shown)


tests/unit_tests/libs/helpers/text_helpers.py
```diff
@@ -0,0 +1,46 @@
...
+        'de': ['führen', 'Stelle', 'heißen', 'konnten', 'schlimm', 'mögen', 'Nähe'],
+    }
+
+    sent_size = 7
+    num_sents = 10
+
+    for lang, words in WORDS.items():
+        sentences = [random.sample(words, sent_size) for _ in range(num_sents)]
+
+        nr_words, word_dist, nr_words_dist = analyze_sentences(' '.join(sent) for sent in sentences)
```
btseytlin, line 24: All good. Would be better if `analyze_sentences` had a more intuitive name matching what it does

btseytlin (approved):

tests/unit_tests/libs/helpers/text_helpers.py
maximlopin: it kinds of does too many things to name it something more specific than "analyze". If you have a name suggestion, we can change it later.

PR #44, opened 2020-07-06
2.0 predict, test, learn interfaces

George3d6:
@torrmal 

This is the new predict/learn interface, please validat it. Basically:

* Kept ts arguments as we don't support them as part of the datasource thus far and can't use dataskillet yet
* Removed `when`, `when_data` can be a dict, more consistent with `learn` having a single `from_data` argument
* Removed a bunch of arguments that were not used
* Removed a bunch of arguments + their usage in mindsdb since they were no longer needed
* Renamed `unstable_parameters_dict` to `advanced_args`... has a nicer ring to it, much more presentable. Telling users they are highly skilled rather than they are playing with fire :P

Anyone else feel free to review, but since this is altering the core functionality of mindsdb_native I'd like @torrmal 's opinion before merging.


mindsdb_native/libs/backends/lightwood.py
```diff
@@ -200,27 +197,25 @@ def train(self):
...
-                self.predictor.learn(from_data=train_df, test_data=test_df, callback_on_iter=self.callback_on_iter, eval_every_x_epochs=eval_every_x_epochs)
-            else:
-                self.predictor.learn(from_data=train_df, test_data=test_df, stop_training_after_seconds=self.transaction.lmd['stop_training_in_x_seconds'], callback_on_iter=self.callback_on_iter, eval_every_x_epochs=eval_every_x_epochs)
+        # Within some limits
+        if eval_every_x_epochs > 200:
+            eval_every_x_epochs = 200
+        if eval_every_x_epochs < 3:
+            eval_every_x_epochs = 3
+
+        logging.getLogger().setLevel(logging.DEBUG)
```
btseytlin, line 212: Some leftover debugging stuff it seems

mindsdb_native/libs/controllers/predictor.py
```diff
@@ -245,7 +236,7 @@ def learn(self,
                 for k in ['data_preparation', 'rebuild_model', 'data_source', 'type', 'columns_to_ignore', 'sample_margin_of_error', 'sample_confidence_level', 'stop_training_in_x_seconds']:
                     if old_lmd[k] is not None: light_transaction_metadata[k] = old_lmd[k]
 
-                for k in ['from_data', 'test_from_data']:
+                for k in ['from_data']:
```
btseytlin: I guess the loop can be removed

mindsdb_native/libs/backends/lightwood.py
George3d6: Nope, there for a reason related to lightwood logging, can't remember what but I'm fairly sure it's relevant or at least not there by mistake at the time it was placed

mindsdb_native/libs/controllers/predictor.py
George3d6: done

btseytlin (approved):
Looks good

mindsdb_native/libs/controllers/predictor.py
```diff
@@ -298,19 +289,24 @@ def predict(self,
...
                 self.log.error(error_msg)
                 raise ValueError(error_msg)
 
             transaction_type = TRANSACTION_PREDICT
-            when_ds = None if when_data is None else getDS(when_data)
-
-            # lets turn into lists: when
-            when = [when] if isinstance(when, dict) else when if when is not None else []
+            when_ds = None
+            when = None
```
maximlopin, line 302: `when` can now end up being `None`.
It's passed to `hmd['model_when_conditions']` and in `data_extractor.py:20` it's treated as an iterable without checking for `None`.

maximlopin (changes requested):

mindsdb_native/libs/controllers/predictor.py
George3d6: No it's not:

```
            elif self.transaction.hmd['model_when_conditions'] is not None:

                # if no data frame yet, make one
                df = self._get_data_frame_from_when_conditions()
```

George3d6: As in, `_get_data_frame_from_when_conditions` is only ever called *after* we figure out `model_when_conditions` is not `None`

George3d6: Buuut actually, good point, if `when` is always a list, why make it `None`, maybe I'm gonna break something. let me make it a list just in case :)

George3d6: Ok, I ended up keeping the default value of `when` as `None` but I simplified some of the data extractor logic so that it's more obvious that it doesn't cause issues.

I also realized I could now make `when_data` a mandatory argument => can avoid some check for the case where both `when` and `when_data` are empty.

mindsdb_native/libs/backends/lightwood.py
torrmal: is the documentation for this updated?

PR #47, opened 2020-07-07
Mongodb datasource

maximlopin:
https://github.com/mindsdb/mindsdb_native/issues/38


optional_requirements_extra_data_sources.txt
```diff
@@ -1,2 +1,3 @@
 boto3 >= 1.9.0
 pg8000 >= 1.15.3
+pymongo >= 3.10.1
```
George3d6, line 3: Maybe allow for older version here, e.g. `3.0.2`

mindsdb_native/libs/data_sources/mongodb_ds.py
```diff
@@ -0,0 +1,25 @@
+import pandas as pd
+from pymongo import MongoClient
+
+from mindsdb_native.libs.data_types.data_source import DataSource
+
+
+class MongoDS(DataSource):
+    def _setup(self, host='localhost', port=27017, user='admin', password='123',
```
George3d6: No point in having a default value for `collection`

Also, it should include some sort of filtering option.

Ideally though, this should just have a `query` parameter that allows the user to user a standard json that one uses when querying mongo.

Something like:

`MongoDs(query='{...mongo query json string}')`

And, because I think the query can include the `collection` inside of it (but correct me if I'm wrong) we can have the `collection` and `query` argument be optional, but at least one of them should always be provided.

btseytlin (approved):

George3d6 (changes requested):

George3d6 (approved):

PR #49, opened 2020-07-07
Add data deduplication and refactor data cleaner slightly

### LINKED ISSUE #16: Deduplicate data
Long discussion, see: https://github.com/mindsdb/mindsdb/issues/414 and the PR resulting from that.

Will probably just re-do it from scratch.

### LINKED ISSUE #48: Only store train/val/test indices instead of copying dataframes
In DataSplitter we do this:
```
self.transaction.input_data.train_df = self.transaction.input_data.data_frame.iloc[train_indexes[KEY_NO_GROUP_BY]].copy()
self.transaction.input_data.test_df = self.transaction.input_data.data_frame.iloc[test_indexes[KEY_NO_GROUP_BY]].copy()
self.transaction.input_data.validation_df = self.transaction.input_data.data_frame.iloc[validation_indexes[KEY_NO_GROUP_BY]].copy()
```

While in fact we can just store the indices, while still using one dataframe under the hood.

This should improve memory consumption for large datasets and allow to simplify code in DataTransformer and following phases

btseytlin:
Resolves: https://github.com/mindsdb/mindsdb_native/issues/16 and https://github.com/mindsdb/mindsdb_native/issues/48

Bonus points: 
* DataCleaner operations are now inplace
* ~~train/test/val dataframes are not copied around~~ turned out to be a bad idea

Everything should work slightly faster!


mindsdb_native/libs/data_types/transaction_data.py
```diff
@@ -4,13 +4,26 @@
...
 
         self._sample_df = None
 
+        self.train_idx = None
+        self.validation_idx = None
+        self.test_idx = None
+
+    @property
+    def train_df(self):
+        return self.data_frame.loc[self.train_idx]
```
George3d6: This change seems like a beyond horrible idea to me.

Index Accessing in pandas is slow and these will be accessed .... hundreds, thousands, millions of times.

This should be removed at all cost, I would be fine with it *IF* it had a cache but even then it seems like a fairly pointless change and I'd rather not change the code unless there's a good reason to.

But maybe I'm wrong here.

mindsdb_native/libs/data_types/transaction_data.py
```diff
@@ -4,13 +4,26 @@
...
+
+    @property
+    def test_df(self):
+        return self.data_frame.loc[self.test_idx]
+
+    @property
+    def validation_df(self):
+        return self.data_frame.loc[self.validation_idx]
```
George3d6: Same as with `def train_df`

mindsdb_native/libs/phases/data_cleaner/data_cleaner.py
```diff
@@ -3,28 +3,32 @@
 
 
 class DataCleaner(BaseModule):
-    def _cleanup_w_missing_targets(self, df):
+    def get_empty_columns(self, df):
```
George3d6: "private" methods should use `_` in front, at least that's the style we went with

mindsdb_native/libs/phases/data_cleaner/data_cleaner.py
```diff
@@ -3,28 +3,32 @@
...
+        empty_columns = []
+        for col_name in df.columns.values:
+            if len(df[col_name].dropna()) < 1:
+                empty_columns.append(col_name)
+                self.log.warning(f'Column "{col_name}" is empty ! We\'ll go ahead and ignore it, please make sure you gave mindsdb the correct data.')
+
+        return empty_columns
+
+    def remove_missing_targets(self, df):
```
George3d6: "private" methods should use `_` in front, at least that's the style we went with

George3d6 (changes requested):

mindsdb_native/libs/phases/data_transformer/data_transformer.py
```diff
@@ -1,88 +1,99 @@
...
+
+        input_data.data_frame.drop(duplicate_idx, inplace=True)
+        if input_data.train_idx is not None:
+            input_data.train_idx = np.array([i for i in input_data.train_idx if i not in duplicate_idx])
+            input_data.validation_idx = np.array([i for i in input_data.validation_idx if i not in duplicate_idx])
+            input_data.test_idx = np.array([i for i in input_data.test_idx if i not in duplicate_idx])
+
+        no_dropped = sum(duplicated)
+        if no_dropped > 0:
+            self.log.warning(f'Dropped {no_dropped} duplicate rows.')
```
George3d6: In principle the logic here works, might even work if you switch back to having dataframes instead of a collection of indexes in the input_data.

George3d6 (changes requested):

mindsdb_native/libs/phases/data_transformer/data_transformer.py
btseytlin: Hmmm, true, actually if `train_df` and other split dataframes preserve the initial index it can work

George3d6 (approved):

mindsdb_native/libs/phases/data_cleaner/data_cleaner.py
```diff
@@ -1,30 +1,40 @@
...
+
+    def _remove_missing_targets(self, df):
+        initial_len = len(df)
+        df.dropna(subset=self.transaction.lmd['predict_columns'], inplace=True)
+        no_dropped = initial_len - len(df)
+        if no_dropped > 0:
+            self.log.warning(
+                f'Dropped {no_dropped} rows because they had null values in one or more of the columns that we are trying to predict. Please always provide non-null values in the columns you want to predict !')
+
+    def _remove_duplicate_rows(self, df):
```
maximlopin, line 22: Everything looks good, but this issue that I mentioned in the old deduplication PR still exists:
> Also, what if my data is thousands rows of random input to an AND gate? It will only keep 4 rows of data, so user should be `log.warned` about deduplication and must have an option to disable it when calling `Predictor.learn`.

Maybe we can at least add a flag in config `DEDUPLICATE_ROWS = True`?

George3d6: I agree with this @btseytlin please just add a `deduplicate` flag in the `advanced_args`, set it to `True` by default but users can change it to `False`, besides that, please fix merge conflicts and go ahead and merge.

PR #51, opened 2020-07-07
Evaluate regression score as pct of true values in confidence intervals

btseytlin:


mindsdb_native/libs/helpers/general_helpers.py
```diff
@@ -134,31 +136,41 @@ def get_value_bucket(value, buckets, col_stats, hmd=None):
...
+            within_interval += 1
+    return within_interval/len(true_values)
+
+
+def evaluate_classification_accuracy(column, predictions, true_values):
+    pred_values = predictions[column]
+    return balanced_accuracy_score(true_values, pred_values)
+
+
+def evaluate_accuracy(predictions, data_frame, col_stats, output_columns, hmd=None):
```
maximlopin: `hmd` is never used

mindsdb_native/libs/helpers/general_helpers.py
```diff
@@ -134,31 +136,41 @@ def get_value_bucket(value, buckets, col_stats, hmd=None):
...
+    evaluator_from_dtype = {
```
maximlopin: I am not sure if it's safe to assume that the data type is either numerical or categorical. If it is, then everything looks perfect.

mindsdb_native/libs/helpers/general_helpers.py
George3d6: `hmd` is there because for *certain* data type it might be needed in the future (e.g. images and text, where the hmd will contain a comparative model to actually evaluate the accuracy), so keep it in there please.

George3d6: @btseytlin this contains a horrible bug not caught by tests, I believe, in that it will crash for everything that isn't NUMERIC or CATEGORICAL.

Which, at the moment, should just be `DATE` types I believe, but for when the backend is set to `ludwig` it could also include other data type. *Please* try to stick to the previous logic, I like the enum here, but maybe add a function `evaluate_generic` for when the data type is not in the enum.

Do not, currently, add tests for all 6 supported output types, even though they would have indeed caught this, but we should add that sooner rather than later, I agree with what I'm 99% sure will be your opinion there.

George3d6: Or maybe do add the tests, if you think it's appropriate for this and quick, I  certainly will make an issue to add them sometimes this week.

mindsdb_native/libs/helpers/general_helpers.py
```diff
@@ -134,31 +136,41 @@ def get_value_bucket(value, buckets, col_stats, hmd=None):
...
+        DATA_TYPES.NUMERIC: evaluate_regression_accuracy,
+        DATA_TYPES.CATEGORICAL: evaluate_classification_accuracy,
+    }
+
+    column_scores = []
+    for column in output_columns:
+        col_type = col_stats[column]['typing']['data_type']
+        evaluator = evaluator_from_dtype[col_type]
```
George3d6: Please replace with an if-else, there's no reason *not* to use an if-else here, it's smaller than a lookup and easier to read, especially since every branch contains a single line.

George3d6 (changes requested):

Issue #50, opened 2020-07-07
Split Transaction into subclasses


btseytlin:
Currenty whenever `Predictor.learn`, `.predict`, `.test` or `analyse_dataset` are called, a `Transaction` object is created.
However for each of these operations it behaves differently. 
This is implemented using methods like `Transaction._execute_learn`, Transaction._execute_analyze` and so on.

It would be cleaner to make subclasses of `Transaction` for each operation.

PR #53, opened 2020-07-07
Split transaction

maximlopin:
https://github.com/mindsdb/mindsdb_native/issues/50


George3d6 (approved):
I'm not 100% sure I like this design @maximlopin , but I can see why you might think it's "cleaner" in that the if/else chain that calls a function is basically replaced by:

```
TYPE_TO_CLASS = {
            TRANSACTION_BAD_QUERY: BadTransaction,
            TRANSACTION_LEARN: LearnTransaction,
            TRANSACTION_ANALYSE: AnalyseTransaction,
            TRANSACTION_PREDICT: PredictTransaction
        }
```

I'm not sure this is better, but I'll grant that there is a part of my OCD that is more appeased by this.

So for now, go ahead and merge, I think it looks nice, as mentioned though, let's avoid this specific type of changes in the future, since they don't add much in terms of speed, functionality, ease of debugging or simplified logic. 

They are just replacing one abstraction for another for the aesthetics and I agree the aesthetics here are nicer, but I don't see a semi-objective stick in the ground to judge that by.

But, again, I'm fairly happy to go ahead and merge this one, just a thing to keep in mind for later :)

PR #55, opened 2020-07-08
ClickHouse test table type changed to MergeTree

StpMax:


tests/integration_tests/data_sources/test_clickhouse_ds.py
```diff
@@ -11,20 +11,25 @@ def test_clickhouse_ds():
...
-                  data="""INSERT INTO test.mock VALUES ('c',3,[3,1,2])""")
+    queries = [
+        'CREATE DATABASE IF NOT EXISTS test',
+        'DROP TABLE IF EXISTS test.mock',
+        '''
+            CREATE TABLE test.mock(
+                col1 String
+                ,col2 Int64
+                ,col3 Array(UInt8)
+            ) ENGINE = MergeTree()
```
George3d6, line 22: Hmh, not sure about removing the `Memory` engine tests... but then again, probably superflous, we can rely on the database to work internally + `MergeTree` is the most widely used. So I'll call this good, go ahead and merge

George3d6 (approved):

PR #59, opened 2020-07-10
Select text encoder

maximlopin:
1. selects `lightwood.encoders.text.short.TextAutoEncoder` (which is added in https://github.com/mindsdb/lightwood/pull/198) in case of `data_subtype=DATA_SUBTYPE.SHORT`

2. Do testing in travis using lightwood/staging now


George3d6 (changes requested):
@maximlopin before we review this, can you add a "switch" to the mindsdb install to somehow run the tests in staging using lightwood's staging branch. Otherwise this parallel dev thing will fail horribly.

We need to make sure that this is a switch so that we can run the tests without it when we do the staging -> stable PR (Or we can just do staging -> something_else_stable -> stable if it's very hard to make it a switch)

Feel free to create an issue with that + work on the PR

Other than that, will try to review the changes + make sure the tests pass for me locally on the morrow or more likely Monday.

mindsdb_native/libs/backends/lightwood.py
```diff
@@ -132,6 +130,9 @@ def _create_lightwood_config(self):
                 'type': lightwood_data_type
             }
 
+            if data_subtype == DATA_SUBTYPES.SHORT:
+                col_config['encoder_class'] = lightwood.encoders.text.short.TextAutoEncoder
```
George3d6: ShortTextEncoder

requirements.txt
```diff
@@ -1,5 +1,5 @@
 psutil>=5.7.0
-lightwood >= 0.27.1
+lightwood >= 0.30
```
George3d6: `0.29.0` and wait for the stable build to deploy before changing the version + switch tests to staging

George3d6 (changes requested):

George3d6 (approved):
Perfect !

PR #60, opened 2020-07-11
Refactor datasource

maximlopin:
- remove redundant `:param:` and `:return:` comments from docstrings
- add comments in `DataSource._cleanup` and `DataSource._setup` informing that they can be overriden
- make `DATA_TYPES_SUBTYPES ` a `dict`
- explicitly call `__getattr__`, `__getitem__`, `__setitem__` in `DataSource` when it maps them to the `DataFrame`
- fix a few for loops so it's easier to follow the logic
- remove test under `__name__ == '__main__'` in  `S3DS` datasource


George3d6 (changes requested):
> add comments in DataSource._cleanup and DataSource._setup informing that they can be overriden

As discussed, just stick to `pass` and don't add the `Can be overriden by subclasses`, that should be the default for any method we have

> explicitly call __getattr__, __getitem__, __setitem__ in DataSource when it maps them to the DataFrame
Why ?

The rest of the changes look fine

maximlopin:
@George3d6 `__getitem__`-like methods are at level of abstraction below the `[]` operator, so calling `__getitem__` which is at the same level of abstraction makes more sense and is more explicit.

George3d6:
Fair point. Go ahead and merge.

George3d6 (approved):

PR #62, opened 2020-07-13
Changes to allow for easier testing

George3d6:
* Changed the `test` method to allow passing the full `.explanation` property to the scoring function (in order allow for e.g. evaluating numerical ranges or evaluation based on the actual predictions + confidences)
* Change `learn` in order to return the validation dataset

Note: Not to be merged just yet, but feel free to review if you have the time.

### DIFF
--- a/mindsdb_native/libs/controllers/predictor.py
+++ b/mindsdb_native/libs/controllers/predictor.py
@@ -243,7 +243,7 @@ class Predictor:
                 if old_hmd['from_data'] is not None:
                     heavy_transaction_metadata['from_data'] = old_hmd['from_data']
 
-            self.transaction = transaction = LearnTransaction(session=self,
+            self.transaction = LearnTransaction(session=self,
                         light_transaction_metadata=light_transaction_metadata,
                         heavy_transaction_metadata=heavy_transaction_metadata,
                         logger=self.log)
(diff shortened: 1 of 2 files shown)


btseytlin:
How about we just return the whole transaction? Then we can inspect the validation df and other info too

Also this Pr is targeting stable. Don't know if intended!

George3d6:
@btseytlin not intended.

Also, returning the whole transaction or making the transaction a visible property of the predictor might indeed be better.

I think, for now, I will also add the `validate` method based on our discussion.

btseytlin (approved):

paxcema (approved):

mindsdb_native/libs/controllers/predictor.py
```diff
@@ -242,10 +243,11 @@ def learn(self,
                 if old_hmd['from_data'] is not None:
                     heavy_transaction_metadata['from_data'] = old_hmd['from_data']
 
-            LearnTransaction(session=self,
+            self.transaction = transaction = LearnTransaction(session=self,
```
maximlopin: typo (`self.transaction = transaction = ...`)

maximlopin (approved):

mindsdb_native/libs/controllers/predictor.py
George3d6: Fixed :)

Issue #70, opened 2020-07-16
DataCleaner issues with ignored column


George3d6:
To replicate clone `mindsdb_examples` and run https://github.com/mindsdb/mindsdb-examples/blob/master/benchmarks/cifar_100/video_tutorial.py

However has time to look at this tomorrow please do, if it's a bug that's not specific to just this niche example let's hotfix it into `stable` (assuming it shows up on stable)

btseytlin:
I will take a look at this

btseytlin:
Here's the fix:
https://github.com/mindsdb/mindsdb_native/pull/72

@George3d6 I suggest we merge it to staging for two reasons: 1. staging to stable release is soon next week and 2. some of the tests I have written for this fix are dependant on a thing introduced in staging (transaction being accessible).

We could merge the tiny fix itself separately into stable, and tests separately into staging of course.

btseytlin:
@maximlopin after the fix the example fails with the following error:
```
ValueError: decode is only defined for combine="concat"
```

It seems to me that mindsdb_native uses the `ShortTextEncoder` here for the output variable `class`, for some reason. Could you please take a look at this?

George3d6:
@btseytlin well pointed, it's because the sample is really weird, making an issue.

Likely not to come up in a real situation, but still important.

PR #72, opened 2020-07-17
Fix data cleaner running twice

btseytlin:
Resolves: https://github.com/mindsdb/mindsdb_native/issues/70

Turns out the problem was that DataCleaner ran twice during LearnTransaction

### DIFF
--- a/mindsdb_native/libs/controllers/transaction.py
+++ b/mindsdb_native/libs/controllers/transaction.py
@@ -152,10 +152,8 @@ class LearnTransaction(Transaction):
                                     input_data=self.input_data)
             self.save_metadata()
 
-            self._call_phase_module(module_name='DataCleaner')
-            self.save_metadata()
-
             self._call_phase_module(module_name='DataSplitter')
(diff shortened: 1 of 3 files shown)


George3d6 (changes requested):
2 things:

a) I'm 80% sure there was a reason the data cleaner ran twice (i.e. some things could only be done after the type deduction and data analysis ran). Please make sure that this is *not* the case if you want to merge it this was, otherwise please go back to the previous "stag based" data cleaner design (since having two separate phases here would be overkill) ... or have two separate phases, if you *really* must.

b) Since this is affecting stable please make the PR into `stable` and add a minor version bump (to, say `2.0.1`) once that is approved go ahead and merge that and auto-merge stable into staging.

btseytlin:
The issue has been hotfixed on stable, but now I will look for a good design longer term solution to it in this PR

George3d6:
I believe the current fix should be good enough, I will review it again just in case.

Backlog any refactoring due to this, not relevant for now.

PR #74, opened 2020-07-17
Hotfix data cleaner on stable

btseytlin:
Hotfixes: https://github.com/mindsdb/mindsdb_native/issues/70


George3d6 (changes requested):
Please also add the missing columns tests you had in the other PR and change  `2.00.1` to `2.0.1` (will get picked up by PyPI just fine, it ignores leading 0s).

Barring that, this is good to merge.

btseytlin:
I am afraid the tests are not possible without changes from staging, so they will arrive with the next release of staging

Issue #68, opened 2020-07-16
Increase sampling threshold and ensure learn & analyse_dataset consistency


George3d6:
* Increase the sampling threshold such that no sampling is done for the statistical analysis up to, say, 10,000 rows. Shouldn't be required on small datasets.
* Make sure the default sampling settings for the Analyse phase are consistent between `learn` and `analyse_dataset`, such that the resulting analysis by calling them without providing that argument is consistent between the two.

PR #75, opened 2020-07-17
Raise tresholds after which sampling and lightwood cache disabling are enabled

btseytlin:
Resolves: https://github.com/mindsdb/mindsdb_native/issues/68

Sampling for analysis now only enabled if the dataset is >= 30% of available RAM, same for disabling lightwood cache.
Sampling for training now only enabled if the dataset is >= 50% of available RAM.

### DIFF
--- a/mindsdb_native/libs/controllers/predictor.py
+++ b/mindsdb_native/libs/controllers/predictor.py
@@ -23,9 +23,9 @@ def _get_memory_optimizations(df):
 
     mem_usage_ratio = df_memory/total_memory
 
-    sample_for_analysis = True if mem_usage_ratio >= 0.05 else False
-    sample_for_training = True if mem_usage_ratio >= 0.15 else False
-    disable_lightwood_transform_cache = True if mem_usage_ratio >= 0.15 else False
+    sample_for_analysis = True if mem_usage_ratio >= 0.3 else False
+    sample_for_training = True if mem_usage_ratio >= 0.5 else False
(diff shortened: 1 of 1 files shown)


maximlopin (approved):

mindsdb_native/libs/controllers/predictor.py
```diff
@@ -23,9 +23,9 @@ def _get_memory_optimizations(df):
 
     mem_usage_ratio = df_memory/total_memory
 
-    sample_for_analysis = True if mem_usage_ratio >= 0.05 else False
-    sample_for_training = True if mem_usage_ratio >= 0.15 else False
-    disable_lightwood_transform_cache = True if mem_usage_ratio >= 0.15 else False
+    sample_for_analysis = True if mem_usage_ratio >= 0.3 else False
```
George3d6, line 26: I agree that this should have been higher, but I think that, as per issue, this should actually be a function of the nr of rows up to a point (i.e. if we sample out of e.g. 500 rows, sampling might cause issues there). I will see how easy it is to get that logic in, do not merge for now, will merge when that is done.

George3d6 (changes requested):

George3d6:
Part of https://github.com/mindsdb/mindsdb_native/pull/77 with modifications, so closing this for now.

mindsdb_native/libs/controllers/predictor.py
btseytlin: What if you have under 10k rows, but with 10k columns in each row?

Issue #69, opened 2020-07-16
Inconsistencies between the test accuracy and the confusion matrix.


George3d6:
Given the following dataset:

[Plackett-Burman_Coffee-July.xlsx](https://github.com/mindsdb/mindsdb_native/files/4934830/Plackett-Burman_Coffee-July.xlsx)

and trying to predit: Coffe_Malt,Chocolat,Gold,Medium_Barley,Dark_Barley,Dandelion,Beets,Chicory_Roots,Figs,Dates,Cacao

There seem to be some inconsistencies between the confusion matrixes and the test accuracy (i.e. the one on the validation dataset) in scout.

Please make sure that the confusion matrix is computed on the validation dataset (the one we also use for the validation_accuracy and for the bayesian confidence models).

Barring that being an issue, please investigate if this might just boil down to an issue with scout and if so followup with @Ricram2

Ricram2:
It seems like an issue on multiple ends. (please check on your end) 
This is one of the responses 

```
confusion_matrix: {matrix: [[4]], predicted: ["0"], real: ["0"]}
matrix: [[4]]
0: [4]
0: 4
predicted: ["0"]
0: "0"
real: ["0"]
0: "0"
```

This is what we show:

![image](https://user-images.githubusercontent.com/24927430/87736400-a4638580-c79d-11ea-95ad-e7a31df8ecff.png)

In this case, the CM is actually doing its job on expressing the result. 

On the other hand in the same analysis there's responses like:


```column_name: "Chocolat"
confusion_matrix: {matrix: [[3, 0], [0, 1]], predicted: ["0", "1"], real: ["0", "1"]}
matrix: [[3, 0], [0, 1]]
0: [3, 0]
1: [0, 1]
predicted: ["0", "1"]
0: "0"
1: "1"
real: ["0", "1"]
0: "0"
1: "1"
```

Where we show:

![image](https://user-images.githubusercontent.com/24927430/87736684-70d52b00-c79e-11ea-9ab0-304df23c3310.png)

And it is clearly not ok. see the issue here: 
https://github.com/mindsdb/mindsdb_gui/issues/476

PR #77, opened 2020-07-17
Various hot~ish fixes

George3d6:
I won't rush any of these into stable since they are really only issues on very tiny datasets.

* Sampling setting modifications that will sample a bit less aggressively + sample more aggressively for large dataframes for the analysis part. This should get us closer to the old behavior, or at least the part of it we enjoyed
* Selecting categorical types for columns with < 6 values regardless of other constrains
* Added sampling logging to the data analyzer even in the case where we don't sample (where we basically just state we are analyzing the full dataset)
* Added a fix for the confusion matrix for when the model is predicting values that aren't within the real data (can happen on very tiny datasets or on datasets where the output is categorical and very dense)


George3d6:
This resolves #69 and #68

mindsdb_native/libs/helpers/probabilistic_validator.py
```diff
@@ -180,7 +180,7 @@ def get_accuracy_stats(self):
             ,'accuracies': list(bucket_accuracy.values())
         }
 
-        labels= list(set(self.real_values_bucketized))
+        labels= list(set([*self.real_values_bucketized, *self.normal_predictions_bucketized]))
```
paxcema, line 183: I am not sure if I understand what this is doing. Set transformation is done to avoid duplicate labels?

paxcema (approved):

mindsdb_native/libs/helpers/probabilistic_validator.py
George3d6: Set transformation is done to avoid duplicates, yes. We are adding both the real values and the predicted values since in some cases, on small datasets, there's a chance for certain values to show up in training but no in the validation dataset (so they are only peresent in the predicted values) => incorrect confusion matrix.

Issue #73, opened 2020-07-17
Don't classify single-words as text


George3d6:
Currently, single-words can sometimes be classified as text.

I think this should happen close to never in real data, and it might lead to some situations where we classify categories as text.

For now, if a sentence contains a single word, can you go ahead and just assume it's a category much like we did before.

Note: Not a bug per se, I know I asked for this initially, I just now realized the probability of cock-ups there is great and the reward tiny.

George3d6:
Potentially misguided request... let's discuss this.

PR #78, opened 2020-07-18
single words are categories

maximlopin:
Resolves https://github.com/mindsdb/mindsdb_native/issues/73
Added special case when all column values are **single** words then the data type is categorical.


George3d6:
Seems to address the issue, very simple change so will just go ahead and merge such that we have it for the release.

PR #79, opened 2020-07-20
Tests for foreign keys being ignored, ignore_columns param working

btseytlin:
Tests for this hotfix: https://github.com/mindsdb/mindsdb_native/pull/74

As a bonus, added `--show-capture=no` default option to pytest. This hides log and stdout output.
If you need log output in tests, you can override it by passing `--show-capture=yes`:
https://docs.pytest.org/en/stable/logging.html

### DIFF
--- a/tests/unit_tests/libs/controllers/test_predictor.py
+++ b/tests/unit_tests/libs/controllers/test_predictor.py
@@ -155,6 +155,39 @@ class TestPredictor:
         assert 'do_use' in transaction.input_data.train_df.columns
         assert 'ignore_this' not in transaction.input_data.train_df.columns
 
+    def test_ignore_foreign_keys(self):
+        input_dataframe = pd.DataFrame({
+            'do_use': list(range(100)),
+            'numeric_id': list(range(100)),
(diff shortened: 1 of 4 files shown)


tests/unit_tests/libs/controllers/test_predictor.py
```diff
@@ -205,7 +258,7 @@ def test_predictor_deduplicate_data(self):
 
         model_data = F.get_model_data('test_drop_duplicates')
 
-        # Ensure duplicate row was not used for training
+        # Duplicate row was used for analysis and training
 
         assert model_data['data_preparation']['total_row_count'] == n_points+1
```
maximlopin, line 263: why is it `n_points + 1`?

btseytlin: @maximlopin there were `n_points` generated, and then one was duplicated, so there are `n_points+1` in the data frame.

This is the second part of the test. It checks that if deduplication is disabled the duplicated row is used for operations.

maximlopin (approved):

PR #82, opened 2020-07-21
Release 2.1.0

btseytlin:
* Always use staging lightwood when running CI tests
* Use lightwood ShortTextEncoder for short text
* Interface for validating trained predictors
* Sample for analysis only if a dataframe has over 10k rows
* Transaction is accessible as `Predictor.transaction` after running `.learn` 
* DataSource names
* Single words are always categorical, not short text
* Things that have few unique values are also always categorical
* Fixed bug where unit tests couldn't be run without installing optional dependencies


btseytlin (commented):
Reviewed, all seems to be good

StpMax (approved):

George3d6 (approved):

paxcema (approved):

mindsdb_native/libs/controllers/predictor.py
```diff
@@ -23,9 +23,9 @@ def _get_memory_optimizations(df):
 
     mem_usage_ratio = df_memory/total_memory
 
-    sample_for_analysis = True if mem_usage_ratio >= 0.05 else False
-    sample_for_training = True if mem_usage_ratio >= 0.15 else False
-    disable_lightwood_transform_cache = True if mem_usage_ratio >= 0.15 else False
+    sample_for_analysis = True if mem_usage_ratio >= 0.1 or len(df) > pow(10,4) else False
```
maximlopin, line 26: replacing `len(df)` with `df.shape[0] * df.shape[1]` makes more sense

maximlopin (approved):

mindsdb_native/libs/controllers/predictor.py
George3d6: @maximlopin if you can, please make this change, it should indeed workout better to take into account length + nr columns.

Also, realistically, the training sampling and disabling the cache should probably happen lightwood-side.

maximlopin: https://github.com/mindsdb/mindsdb_native/pull/84

PR #83, opened 2020-07-21
set ds 'table' param to None by default

StpMax:
In mindsb 'table' arg not used when datasources creates, so it better to set it to None by default (except MongoDS, will think about it later)


maximlopin:
in `_setup `it does `SELECT * from {table}` though, why not make `table `a positional required argument?

StpMax:
As i see 'table' arg used just in mindsdb_native tests. And in all tests it used as named arg. In mindsb all interaction with datasources goes via 'query' arg, and at this moment to save/load it need to store additional `'table': None` arg.

PR #84, opened 2020-07-22
enable analysis sampling based on nr_cols*nr_rows

maximlopin:


George3d6:
pow(10, 4) should probably be, say, 3*pow(10, 4)

Basically, increase the number a bit since the left hand side is now much bigger.

George3d6 (approved):

paxcema (approved):

PR #85, opened 2020-07-23
Use test dataset for confidence model

George3d6:
* Using the test data frame for the confidence model
* Removed validation dataset label duplication since that was only there for the confidence model


paxcema (approved):

mindsdb_native/libs/phases/data_transformer/data_transformer.py
```diff
@@ -146,9 +146,7 @@ def run(self, input_data):
                 #print(self.transaction.lmd['weight_map'])
                 column_is_weighted_in_train = column in self.transaction.lmd['weight_map']
 
-                if column_is_weighted_in_train:
-                    dfs = ['input_data.validation_df']
-                else:
+                if not column_is_weighted_in_train:
                     dfs = ['input_data.train_df','input_data.test_df','input_data.validation_df']
```
maximlopin, line 150: `dfs` is only defined in this `if `statement, but is referenced 6 lines below in a `for `loop independently from the `if` statement

mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
```diff
@@ -90,9 +94,9 @@ def run(self):
 
         for col in output_columns:
             pval = ProbabilisticValidator(col_stats=self.transaction.lmd['stats_v2'][col], col_name=col, input_columns=input_columns)
-            predictions_arr = [normal_predictions] + [empty_input_predictions[col] for col in ignorable_input_columns]
+            predictions_arr = [normal_predictions_test] + [empty_input_predictions_test[col] for col in ignorable_input_columns]
```
maximlopin, line 97: aren't these square brackets in the first term making a list whose first item is another list? is it intended?

maximlopin (changes requested):

mindsdb_native/libs/phases/data_transformer/data_transformer.py
George3d6: Oh blimey, that could have been bad... need to figure out how to modify tests for this.

mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
George3d6: yeap, this is indeed, it's an array of arrays of predictions.

maximlopin (approved):

PR #86, opened 2020-07-26
user-provided null values

maximlopin:


George3d6:
@maximlopin this is actually a great idea :+1:

Issue #80, opened 2020-07-20
Data cleaner runs twice to drop foreign keys


btseytlin:
The whole design of mindsdb native is based on sequentially executed phases that process data.

But the `DataCleaner` phase runs twice in `LearnTransaction`, as a special case. This is used as a hack, because:
* Data has to be cleaned before running analysis by dropping empty columns and duplicate rows
* DataAnalysis identifies foreign keys, which by default have to be dropped too. It's wrong to have DataAnalysis drop anything (it's not the responsibility of the phase). So DataCleaner is executed again to drop the foreign keys.

This contradicts the core architecture of mindsdb phases. IIt is very opaque - we only revealed it because of a bug. At some point, we have to rework it, so it's more transparent and contributors can understand what's going on.

My best idea for a new design:
* DataCleaner stays after DataExtractor, because things like dropping duplicates should be done before data analysis
* Dropping foreign keys is moved to DataTransformer
* DataTransformer is moved before DataSplitter. Except for the logic of DataTransformer that operates on separate splits (the reweighting thing), this is moved to DataSplitter.

The good thing is that we already have tests for all the cases covered and the bug that appeared when I touched that code:
https://github.com/mindsdb/mindsdb_native/pull/79

Issue #52, opened 2020-07-07
Make DataTransformer phase run before DataSplitter


btseytlin:
It seems like we could run most if not all of the DataTransformer work before splitting into train/test/validation and the code will be much cleaner.

I am not sure though about the "Unbias dataset for training logic" here:
https://github.com/mindsdb/mindsdb_native/blob/stable/mindsdb_native/libs/phases/data_transformer/data_transformer.py#L126
I don't understand any of it, maybe it's dependant on the splits.

George3d6:
@btseytlin this becomes a problem when you consider the `ludwig` backend requires duplication to be done in all datasets for equal weighting, but the ModelAnalyzer only requries duplciation in the `validation` datasets (though we might change that to something fancier in the future or even just move the `ModelAnalyzer` to lightwood entirely.

Also, we might want to "transform" other things in one of the datasets but no the other

btseytlin:
I can think of some things that we might want to only do on train, but not on test: data augmentation for images.
But most of the work in data transformer is not like this, its like converting string numbers to ints, converting dates to the same format, handling nan values, and so on. It is actually data cleaning work

We could do this:

DataExtractor -> DataAnalyzer -> DataCleaner (with all the cleaning logic that is in Transform) -> ...

DataAnalyzer only depends on DataCleaner in the regard that DataCleaner identifies empty columns and signals to ignore them. This can be done in DataAnalyzer too. And then DataCleaner can use the info of DataAnalysis to "smartly" preprocess data.

What do you think?

George3d6:
We need to merge some of the changes in that PR once's there enough time, couldn't merge as it (reasons for can be seen in notion.io)

PR #91, opened 2020-07-28
Refactor DataCleaner and DataTransformer

btseytlin:
Resolves:
https://github.com/mindsdb/mindsdb_native/issues/80
https://github.com/mindsdb/mindsdb_native/issues/52

### DIFF
--- a/mindsdb_native/libs/backends/lightwood.py
+++ b/mindsdb_native/libs/backends/lightwood.py
@@ -87,10 +87,7 @@ class LightwoodBackend():
         config['input_features'] = []
         config['output_features'] = []
 
-        for col_name in self.transaction.input_data.columns or col_name not in self.transaction.lmd['stats_v2']:
-            if col_name in self.transaction.lmd['columns_to_ignore']:
-                continue
-
+        for col_name in self.transaction.input_data.columns:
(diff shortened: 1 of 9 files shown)


mindsdb_native/libs/backends/lightwood.py
```diff
@@ -87,10 +87,7 @@ def _create_lightwood_config(self):
         config['input_features'] = []
         config['output_features'] = []
 
-        for col_name in self.transaction.input_data.columns or col_name not in self.transaction.lmd['stats_v2']:
-            if col_name in self.transaction.lmd['columns_to_ignore']:
-                continue
-
+        for col_name in self.transaction.input_data.columns:
```
maximlopin, line 90: shouldn't `columns_to_ignore` be skipped?

btseytlin: They are dropped from input data before that

George3d6:
Added my thoughts on this here: https://www.notion.so/mindsdb/Data-Cleaner-Transformer-redesign-evaluation-717af604d6b44606a31c024f6f9a4119

TL;DR

What's happening in the PR makes sense, in this form though it touches a few too many things and it would break some functionality, so I'd rather modify it slightly and merge it later.

Will close for now and re-open in the future when we have time to address those things.

PR #93, opened 2020-07-31
user-provided-subtypes

maximlopin:


George3d6:
@maximlopin not sure we want this as an advanced arg for native, since we can just create the datasource as such, but if you feel like it's easier to use this way, feel free to go ahead and merge it :)

Issue #36, opened 2020-07-02
Negative values for column importance.


Ricram2:
**Please describe your issue and how we can replicate it*

Dataset: African Crisis.
Model: 
[afr_year.zip](https://github.com/mindsdb/mindsdb_native/files/4866261/afr_year.zip)

![image](https://user-images.githubusercontent.com/24927430/86401971-b78c3680-bc70-11ea-84d2-b7098518d12e.png)

```
...
   {
            "column_name":"banking_crisis",
            "importance_score":-1.0,
            "data_type":"Categorical",
            "data_type_distribution":{
               "type":"categorical",
               "x":[
                  "Categorical"
               ],
               "y":[
                  1044
               ]
            },
...
```

issue: we are not supposed to have negative values.

maximlopin:
resolved by https://github.com/mindsdb/mindsdb_native/pull/100

PR #99, opened 2020-08-05
New word histogram logic

George3d6:
* Limit words histogram to the top 50 words that aren't stopwords
* Simplify word histogram computation
* Add "other words" to count the rest of the words


maximlopin:
`test_language_analysis` test fails because it doesn't expect stopwords to be removed from the word distribution

maximlopin:
I think https://github.com/mindsdb/mindsdb_native/issues/36 is related to the test failing:
```
>           assert (importance >= 0 and importance <= 10)
E           assert (-1.0 >= 0)
```

PR #100, opened 2020-08-06
change column importance formula

maximlopin:
resolves https://github.com/mindsdb/mindsdb_native/issues/36

**column importance**
equal to 0.5 means removing the column doesn't affect the accuracy
below 0.5 means removing the column increases the accuracy
above 0.5 means removing the column decreases the accuracy

previously the formula was:
`column_importance = (1 - empty_input_accuracy[col] / normal_accuracy)`
examples of values that will result in negative column importance:
```
empty_input_accuracy[col] = 1.0
normal_accuracy = 0.9

column_importance = (1 - empty_input_accuracy[col] / normal_accuracy) = (1 - 1.111) = -0.111
```

### DIFF
--- a/mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
+++ b/mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
@@ -51,7 +51,7 @@ class ModelAnalyzer(BaseModule):
         for col in ignorable_input_columns:
             accuracy_increase = (normal_accuracy - empty_input_accuracy[col])
             # normalize from 0 to 10
-            self.transaction.lmd['column_importances'][col] = 5 * (accuracy_increase + 1)
+            self.transaction.lmd['column_importances'][col] = 10 * max(0, accuracy_increase)
 
         # Run Probabilistic Validator
         overall_accuracy_arr = []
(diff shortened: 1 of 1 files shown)


maximlopin:
New formula:
```
accuracy_increase = (normal_accuracy - empty_input_accuracy[col])
self.transaction.lmd['column_importances'][col] = 5 * (accuracy_increase + 1)
```

**but**

The following formula makes more sense for column **importance**, it only takes to account increase in accuracy when removing a column, and ignores decrease in accuracy:
```
accuracy_increase = (normal_accuracy - empty_input_accuracy[col])
self.transaction.lmd['column_importances'][col] = 10 * max(0, accuracy_increase)
```

George3d6:
I think that the second formula:

```
accuracy_increase = (normal_accuracy - empty_input_accuracy[col])
self.transaction.lmd['column_importances'][col] = 10 * max(0, accuracy_increase)
```

Makes a lot more sense. Since the case where accuracy is higher with a column removed is... nonsensical. Or rather, it's just not something we should care about, it means we have a very bad model anyway and we shouldn't look at these sort of "fine tuning" metrics.

George3d6:
So I'd implement that variant.

PR #108, opened 2020-08-11
user-provided train/test/valid indices

maximlopin:
- add `data_split_indexes` advanced argument
- add test for this
- set `use_gpu` to `False` in tests that call `Predictor.learn`

### DIFF
--- a/mindsdb_native/libs/phases/data_extractor/data_extractor.py
+++ b/mindsdb_native/libs/phases/data_extractor/data_extractor.py
@@ -60,9 +60,14 @@ class DataExtractor(BaseModule):
                 asc_values = [True for i in self.transaction.lmd['model_group_by']] + asc_values
             df = df.sort_values(sort_by, ascending=asc_values)
 
-        elif self.transaction.lmd['type'] == TRANSACTION_LEARN and self.transaction.lmd.get('data_split_indexes') is None:
-            # if its not a time series, randomize the input data and we are learning
-            df = df.sample(frac=1, random_state=len(df))
+        # if its not a time series, randomize the input data and we are learning
+        elif self.transaction.lmd['type'] == TRANSACTION_LEARN:
+            data_split_indexes = self.transaction.lmd.get('data_split_indexes')
+            if data_split_indexes is None:
+                df = df.sample(frac=1, random_state=len(df))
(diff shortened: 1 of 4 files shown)


mindsdb_native/libs/phases/data_extractor/data_extractor.py
```diff
@@ -60,9 +60,14 @@ def _apply_sort_conditions_to_df(self, df):
...
+        # if its not a time series, randomize the input data and we are learning
         elif self.transaction.lmd['type'] == TRANSACTION_LEARN:
-            # if its not a time series, randomize the input data and we are learning
-            df = df.sample(frac=1, random_state=len(df))
+            data_split_indexes = self.transaction.lmd.get('data_split_indexes')
+            if data_split_indexes is None:
+                df = df.sample(frac=1, random_state=len(df))
+            else:
+                for v in data_split_indexes.values():
+                    random.shuffle(v)
```
George3d6: The shuffle isn't actually needed here, since lightwood does it's own shuffling.

It's only relevant for the case in which the user doesn't specify the indexes, since otherwise we risk biasing the dataset since there's often some unforeseen "order" in the way rows are placed in a csv.

However, other than that, shuffling is relevant just for training and lightwood already shuffles the training data at each epoch (and ludwig does the same).

George3d6 (changes requested):

mindsdb_native/libs/phases/data_extractor/data_extractor.py
maximlopin: it used to be just `df = df.sample(frac=1, random_state=len(df))`, which shuffles the `DataFrame`. That's why I kept this logic for the case when `data_split_indexes` is provided. Other phases might have assumptions about how random the data is because that's how it always worked with the "`df = df.sample(frac=1, random_state=len(df))`", it's safer to keep the shuffle, but if you're sure it's ok to remove it, I will

George3d6: I think it's safe to remove, the other phases should have no assumptions about this.

ZoranPandovski:
**CLA Assistant Lite** All Contributors have signed the CLA.

PR #109, opened 2020-08-11
guess probability

maximlopin:
- Add `'guess_probability'` key to `stats_v2[col_name]` when data type of `col_name` is `DATA_TYPES.CATEGORICAL`. This key contains a float - probability of randomly guessing the category.
- Add a test for that


George3d6 (approved):
Is this required for anything *but* the target columns ? If no, I'd guess I'd be better to compute it just for those I think.
Other than that, it's quick and potentially useful, I like it.

maximlopin:
If we only compute it for output columns it will only be in `stats` for a `LearnTransaction`, also it's not expensive to compute and it is kind of a statistic that makes sense to have in `stats`

Issue #120, opened 2020-08-20
ID column is not correctly identified as such


paxcema:
**Your Environment**

* Python version: 3.6.11
* Operating system: Ubuntu 16.04
* Mindsdb native version: 2.3.0
* Additional info if applicable:

**Please describe your issue and how we can replicate it**
For the `automobile_insurance` dataset, column `'Customer'` is an ID column (8220 different values for 8220 rows), however it is not identified as such which results in creating and training an unnecessarily big autoencoder.

maximlopin:
A few thoughts:

Here's the distribution of letters in the `'Customer'` column, we can try doing something with this information to identify it as ID:
`Counter({'7': 3866, '9': 3842, '5': 3842, '3': 3839, '8': 3824, '6': 3801, '4': 3771, '1': 3766, '2': 3688, '0': 3172, 'W': 730, 'M': 708, 'E': 707, 'C': 706, 'F': 706, 'P': 703, 'L': 703, 'Z': 702, 'G': 700, 'N': 700, 'U': 698, 'X': 698, 'B': 694, 'H': 694, 'Q': 693, 'S': 692, 'Y': 689, 'K': 684, 'O': 673, 'V': 672, 'A': 671, 'D': 670, 'R': 665, 'I': 662, 'T': 657, 'J': 650, 'o': 16})`

- `scipy.entorpy` might be useful for measuring "randomness" of the distribution above.
- It's probably better to take the position (index) of each character of ID to account somehow.

maximlopin:
resolved by https://github.com/mindsdb/mindsdb_native/pull/124

PR #124, opened 2020-08-24
improve foreign key detection

maximlopin:
Resolves https://github.com/mindsdb/mindsdb_native/issues/120

### DIFF
--- a/mindsdb_native/libs/helpers/text_helpers.py
+++ b/mindsdb_native/libs/helpers/text_helpers.py
@@ -165,6 +165,9 @@ def _is_foreign_key_name(name):
 def is_foreign_key(data, column_name, data_subtype, other_potential_subtypes):
     data = list(data)
 
+    if len(data) == 0:
+        return False
+
     foregin_key_type = DATA_SUBTYPES.INT in [*other_potential_subtypes, data_subtype]
 
     # Detect UUID
--- a/mindsdb_native/libs/helpers/text_helpers.py
+++ b/mindsdb_native/libs/helpers/text_helpers.py
@@ -192,8 +192,8 @@ def is_foreign_key(data, column_name, data_subtype, other_potential_subtypes):
             randomness_per_index = []
             for i, _ in enumerate(str_data[0]):
                 all_ascii = all(isascii(x[i]) for x in str_data)
-                all_alpha = all(x[i].isalpha() for x in str_data)
-                all_numeric = all(x[i].isnumeric() for x in str_data)
+                all_alpha = all_ascii and all(x[i].isalpha() for x in str_data)
+                all_numeric = all_ascii and all(x[i].isnumeric() for x in str_data)
 
                 if all_alpha:
                     N = 26
--- a/mindsdb_native/libs/helpers/text_helpers.py
+++ b/mindsdb_native/libs/helpers/text_helpers.py
@@ -186,28 +186,17 @@ def is_foreign_key(data, column_name, data_subtype, other_potential_subtypes):
         all_uuid_charset = all(set(str(x)).issubset(uuid_charset) for x in data)
         is_uuid = all_uuid_charset and all_same_length
 
-        if all_same_length:
+        if all_same_length and len(data) == len(set(data)):
             str_data = [str(x) for x in data]
             
             randomness_per_index = []
             for i, _ in enumerate(str_data[0]):
-                all_ascii = all(isascii(x[i]) for x in str_data)
-                all_alpha = all_ascii and all(x[i].isalpha() for x in str_data)
-                all_numeric = all_ascii and all(x[i].isnumeric() for x in str_data)
+                N = len(set(x[i] for x in str_data))
+                S = st.entropy([*Counter(x[i] for x in str_data).values()])
+                randomness_per_index.append(S / np.log(N))
 
-                if all_alpha:
-                    N = 26
-                elif all_numeric:
-                    N = 10
-
-                if all_alpha or all_numeric:
-                    S = st.entropy([*Counter(x[i] for x in str_data).values()])
-                    randomness_per_index.append(S / np.log(N))
-                else:
-                    break
-            else:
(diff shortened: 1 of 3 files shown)


mindsdbadmin:
**CLA Assistant Lite** All Contributors have signed the CLA.

mindsdbadmin:
**CLA Assistant Lite** All Contributors have signed the CLA.

mindsdbadmin:
**CLA Assistant Lite** All Contributors have signed the CLA.

mindsdbadmin:
**CLA Assistant Lite** All Contributors have signed the CLA.

mindsdbadmin:
**CLA Assistant Lite** All Contributors have signed the CLA.

mindsdbadmin:
**CLA Assistant Lite** All Contributors have signed the CLA.

mindsdb_native/libs/helpers/text_helpers.py
```diff
@@ -175,6 +186,29 @@ def is_foreign_key(data, column_name, data_subtype, other_potential_subtypes):
...
+            for i, _ in enumerate(str_data[0]):
+                all_ascii = all(isascii(x[i]) for x in str_data)
+                all_alpha = all_ascii and all(x[i].isalpha() for x in str_data)
+                all_numeric = all_ascii and all(x[i].isnumeric() for x in str_data)
+
+                if all_alpha:
+                    N = 26
+                elif all_numeric:
+                    N = 10
+
```
paxcema: The problematic column in the `automobile_insurance` dataset is actually a mix of alphanumerical characters, so ideally we'd want to check entropy in that case as well.

paxcema: What I mean is that the N should be 36 in that case, if I'm correct.

paxcema (changes requested):

mindsdb_native/libs/helpers/text_helpers.py
```diff
@@ -161,10 +162,20 @@ def _is_foreign_key_name(name):
...
+    Used instead of str.isascii because python 3.6 doesn't have that
+    """
+    return all(ord(c) < 128 for c in string)
+
+
 def is_foreign_key(data, column_name, data_subtype, other_potential_subtypes):
     data = list(data)
 
-    foregin_key_type = DATA_SUBTYPES.INT in other_potential_subtypes or DATA_SUBTYPES.INT == data_subtype
+    if len(data) == 0:
```
George3d6: When is this needed ? The length of the data passed to this should never be == 0

mindsdb_native/libs/helpers/text_helpers.py
```diff
@@ -175,6 +186,29 @@ def is_foreign_key(data, column_name, data_subtype, other_potential_subtypes):
...
+
+                if all_alpha or all_numeric:
+                    S = st.entropy([*Counter(x[i] for x in str_data).values()])
+                    randomness_per_index.append(S / np.log(N))
+                else:
+                    break
+            else:
+                if np.mean(randomness_per_index) > 0.95:
+                    return True
+
```
George3d6: This approach seems overly complex to me. A safe assumption would be that we're bound never to have whitespaces (` `) in a foreign key and that the character density is tiny (e.g. < 50).

In that situation a better approach might be to check for the number of distinct values in the column itself, if that number is ~= the length of the data itself + there are no spaces (i.e. it's probably not text) => foreign key.

@paxcema and @maximlopin what are your thoughts on that ? We certainly need to find some solution here at any rate.

The alternative would be, based on what @paxcema  is saying, to allow for alphanumeric + number and have `N=36`, or even to allow for those + common id characters (e.g. `-` and `_`) and increase N to 38 or 40 or w/e.

Also, since we are now being much more aggressive with foreign key detection we should add a `force_use_column` argument to the `advanced_args` in case we ever mis-diagnose something as a foreign key.

George3d6 (changes requested):

mindsdb_native/libs/helpers/text_helpers.py
maximlopin: this was failing https://github.com/mindsdb/mindsdb_native/blob/stable/tests/unit_tests/libs/phases/type_deductor/test_type_deductor.py#L124

mindsdbadmin:

mindsdb_native/libs/helpers/text_helpers.py
maximlopin: actually this would work for `automobile_insurance` dataset (because it has only alphabetical characters in first 2 indices and others are numeric), but I changed computation for N to `len(set(x[i] for x in str_data))` because it's not restricted to any charset

mindsdb_native/libs/helpers/text_helpers.py
maximlopin: I think we already have logic for disabling handling foreign keys: https://github.com/mindsdb/mindsdb_native/blob/staging/mindsdb_native/libs/phases/type_deductor/type_deductor.py#L322

maximlopin: > In that situation a better approach might be to check for the number of distinct values in the column itself, if that number is ~= the length of the data itself + there are no spaces (i.e. it's probably not text) => foreign key.

I like this idea, but I think it should be implemented in a separate PR

paxcema: I agree with the approach @George3d6 suggests, it feels intuitive as it's similar to the check I did when inspecting the column manually. In any case, this approach seems to work so we should leave the alternative for another PR, as Maxim says.

paxcema (approved):

mindsdb_native/libs/helpers/text_helpers.py
George3d6: That test is then wrong, since this should never be called on an empty column, the column should be removed before in the DataCleaner, remove the test

George3d6 (changes requested):

PR #135, opened 2020-08-25
balanced guess probability

### LINKED ISSUE #132: Use balanced_accuracy to determine "failed" category predicting models
Mindsdb targets balanced_accuracy instead of normal accuracy for models predicting categories, thus the decision to allow a model to fail should be based on the balanced_accuracy and the "better than random" should be determine based on that (might even make the determination easier, since I guess it would just be 1/nr_of_classes, but maybe I'm wrong here)

maximlopin:
Resolves https://github.com/mindsdb/mindsdb_native/issues/132

- use "balanced guess probability" and `balanced_accuracy_score` for detecting failed model
- added a comment with explanation for hash-like foreign key detection


mindsdbadmin:

PR #138, opened 2020-08-26
Confidence using conformal predictors

paxcema:
This PR introduces confidence range estimation through the conformal prediction framework, by using the implementation offered in the [`nonconformist`](https://github.com/donlnz/nonconformist) library (MIT licensed).

As it stands, the `Predictor.predict()` method now exposes a `'conformal_range'` list of `[lower_bound, upper_bound]` prediction ranges for every predicted value. For now, the significance level is at 95%, which means that the range includes the true value with `p = 0.95`. For more details on the technique this [tutorial](https://cml.rhul.ac.uk/copa2017/presentations/CP_Tutorial_2017.pdf) is pretty good.

Quick tests indicate that the out-of-bounds issue, where the value predicted by MindsDB would sometimes fall outside of the prediction range, does not occur with this approach.

Next steps are to:
1. Benchmark this PR with our suite
2. Normalize the ranges by using the self-aware network as a nonconformity measure. In practice, this leads to more precise ranges.
3. Use as a complement with the current approach, by comparing its range and confidence `p` with the one computed by the inductive conformal predictor (`icp`) object.


mindsdbadmin:

mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
```diff
@@ -120,3 +123,40 @@ def run(self):
...
+            icp_class = IcpClassifier
+        else:
+            adapter = ConformalRegressorAdapter
+            nc_function = AbsErrorErrFunc()
+            nc_class = RegressorNc
+            icp_class = IcpRegressor
+
+        model = adapter(self.transaction.model_backend.predictor, fit_params=fit_params)
+        nc = nc_class(model, nc_function)
+        self.transaction.icp = icp_class(nc)
```
George3d6: This should probably be part of the `hmd` and get pickled/unpickled whenever the model is saved, to be available during the `predict` calls, unless it's never used during predict calls (It will work if the mindsdb object is never recreate it, but if you load the predictor anew it won't be saved)

mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
```diff
@@ -120,3 +123,40 @@ def run(self):
...
+        self.transaction.icp.fit(X.values, y.values)
+
+        # calibrate conformal estimator on test set
+        X = self.transaction.input_data.test_df.copy(deep=True)
```
George3d6: nitpick, but maybe switch to `deepcopy` (the python built in function `from copy import deepcopy` since that's what we always use, unless this has an advantage (in which case we should always do it like this)

mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
```diff
@@ -120,3 +123,40 @@ def run(self):
             self.transaction.hmd['probabilistic_validators'][col] = pickle_obj(pval)
 
         self.transaction.lmd['validation_set_accuracy'] = sum(overall_accuracy_arr)/len(overall_accuracy_arr)
+
+        # conformal prediction confidence estimation
+        target = self.transaction.lmd['predict_columns'][0]
+        X = self.transaction.input_data.train_df.copy(deep=True)
```
George3d6: same comment about the copying.

George3d6 (changes requested):

paxcema:
Okay, at this point we should start benchmarking in a formal way (i.e. using our benchmarking suite). However, preliminary results I've gotten show nice performance, comparing favourably against the current approach.

tests/unit_tests/libs/helpers/test_conformal.py
```diff
@@ -0,0 +1,82 @@
...
+from mindsdb import Predictor
+from sklearn.model_selection import train_test_split
+from sklearn.datasets import load_boston, load_iris
+
+
+class TestEvaluateAccuracy(unittest.TestCase):
+
+    def _df_from_x(self, x, columns=None):
+        x = pd.DataFrame(x)
+        if isinstance(columns, type(None)):
```
maximlopin: isn't it just `if columns is None`?

mindsdb_native/libs/controllers/transaction.py
```diff
@@ -282,6 +317,48 @@ def run(self):
...
+                                output_data[f'{predicted_col}_confidence_range'][sample_idx] = list(sample[:, idx])
+                                break
+                        else:
+                            output_data[f'{predicted_col}_confidence'][sample_idx] = 0.9901  # default
+                            bounds = sample[:, 0]
+                            sigma = (bounds[1] - bounds[0])/2
+                            output_data[f'{predicted_col}_confidence_range'][sample_idx] = [bounds[0] - sigma, bounds[1] + sigma]
+
+                else:
+                    if self.lmd['stats_v2'][predicted_col]['typing']['data_subtype'] != 'Tags':
```
maximlopin: better reference the subtype itself:
`if self.lmd['stats_v2'][predicted_col]['typing']['data_subtype'] != DATA_SUBTYPES.TAGS:`

mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
```diff
@@ -152,3 +157,72 @@ def predict_wrapper(*args, **kwargs):
...
+            is_classification = data_type == DATA_TYPES.CATEGORICAL
+            self.transaction.lmd['stats_v2']['is_classification'][target] = is_classification
+
+            fit_params = {'target': target,
+                          'all_columns': self.transaction.lmd['columns'],
+                          'columns_to_ignore': self.transaction.lmd['columns_to_ignore'] +
+                                               [col for col in output_columns if col != target]}
+
+            if is_classification:
+                if data_subtype != 'Tags':
```
maximlopin: `DATA_SUBTYPES.TAGS` instead of 'Tags'

mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
```diff
@@ -152,3 +157,72 @@ def predict_wrapper(*args, **kwargs):
...
+                icp_class = IcpRegressor
+
+            model = adapter(self.transaction.model_backend.predictor, fit_params=fit_params)
+            nc = nc_class(model, nc_function)
+            if is_classification:
+                self.transaction.hmd['icp'][target] = icp_class(nc, smoothing=False)
+            else:
+                self.transaction.hmd['icp'][target] = icp_class(nc)
+
+            if data_subtype != "Tags":
```
maximlopin: `DATA_SUBTYPES.TAGS` instead of 'Tags'

tests/unit_tests/libs/helpers/test_conformal.py
```diff
@@ -0,0 +1,65 @@
+import unittest
+import pandas as pd
+from mindsdb_native import Predictor
+from sklearn.model_selection import train_test_split
+from sklearn.datasets import load_boston, load_iris
+
+
+class TestEvaluateAccuracy(unittest.TestCase):
```
maximlopin: we use `pytest` in mindsdb_native

maximlopin (changes requested):

mindsdb_native/libs/controllers/transaction.py
```diff
@@ -282,6 +317,48 @@ def run(self):
...
                     output_data[f'{predicted_col}_confidence'][row_number] = probability_true_prediction
 
+            # confidence estimation
+            self.lmd['all_conformal_ranges'] = {}
+            for predicted_col in self.lmd['predict_columns']:
+                X = deepcopy(self.input_data.data_frame)
+                for col in self.lmd['columns_to_ignore'] + self.lmd['predict_columns']:
+                    X.pop(col)
+
+                if not self.lmd['stats_v2']['is_classification'][predicted_col]:
```
maximlopin: I don't think it's worth making key `is_classification` becuase `stats_v2[col]['typing']['data_type'] == DATA_TYPES.CATEGORICAL` is simple enough. It works either way though.

maximlopin (approved):

George3d6:
Ideally: dill >= 0.2.9 so that it's the same as what lightwood uses.

mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
```diff
@@ -152,3 +157,70 @@ def predict_wrapper(*args, **kwargs):
...
+                y = X.pop(target).values
+                [X.pop(col) for col in output_columns if col != target]
+
+                if is_classification:
+                    if isinstance(enc.categories_[0][0], str):
+                        cats = enc.categories_[0].tolist()
+                        y = np.array([cats.index(i) for i in y])
+                    y = y.astype(int)
+
+                self.transaction.hmd['icp'][target].calibrate(X.values, y)
```
George3d6, line 228: Is this going to be able to work when the input data contains types such as `text` or `image` or `array` ?

tests/unit_tests/libs/controllers/test_predictor.py
```diff
@@ -305,7 +305,7 @@ def test_explain_prediction(self):
         explanation_new = result[0].explanation['numeric_y']
         assert isinstance(explanation_new['predicted_value'], int)
         assert isinstance(explanation_new['confidence_interval'],list)
-        assert isinstance(explanation_new['confidence_interval'][0],int)
+        assert isinstance(explanation_new['confidence_interval'][0],float)
```
George3d6, line 309: Not sure where to leave this comment, but I fail to see where `confidence_interval` actually ends up being set by the NC model ?

mindsdb_native/libs/controllers/transaction.py
```diff
@@ -282,6 +317,48 @@ def run(self):
 
                     output_data[f'{predicted_col}_confidence'][row_number] = probability_true_prediction
 
+            # confidence estimation
```
George3d6, line 320: It might make more sense for this bit here to be under it's own phase or under the `ModelAnalysis` phase, but there's a bunch of older logic in here for which this is true, so let's ignore this for now.

mindsdb_native/libs/controllers/transaction.py
```diff
@@ -282,6 +317,48 @@ def run(self):
...
+                                output_data[f'{predicted_col}_confidence'][sample_idx] = significance
+                                output_data[f'{predicted_col}_confidence_range'][sample_idx] = list(sample[:, idx])
+                                break
+                        else:
+                            output_data[f'{predicted_col}_confidence'][sample_idx] = 0.9901  # default
+                            bounds = sample[:, 0]
+                            sigma = (bounds[1] - bounds[0])/2
+                            output_data[f'{predicted_col}_confidence_range'][sample_idx] = [bounds[0] - sigma, bounds[1] + sigma]
+
+                else:
```
George3d6: It seems like this should check for == Numerical, otherwise it will just end up failing for outputs that are timeseries, text, images... etc

mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
```diff
@@ -152,3 +157,70 @@ def predict_wrapper(*args, **kwargs):
...
+                else:
+                    fit_params['one_hot_enc'] = None
+                    self.transaction.hmd['label_encoders'][target] = None
+
+                adapter = ConformalClassifierAdapter
+                nc_function = MarginErrFunc()  # better than IPS as we'd need the complete distribution over all classes
+                nc_class = ClassifierNc
+                icp_class = IcpClassifier
+
+            else:
```
George3d6, line 195: It seems like this should check for == Numerical, otherwise it will just end up failing for outputs that are timeseries, text, images... etc

mindsdb_native/libs/controllers/transaction.py
```diff
@@ -282,6 +317,48 @@ def run(self):
...
+                        sample = self.lmd['all_conformal_ranges'][predicted_col][sample_idx, :, :]
+                        for idx in range(sample.shape[1]):
+                            significance = (99 - idx) / 100
+                            diff = sample[1, idx] - sample[0, idx]
+                            if diff <= tolerance:
+                                output_data[f'{predicted_col}_confidence'][sample_idx] = significance
+                                output_data[f'{predicted_col}_confidence_range'][sample_idx] = list(sample[:, idx])
+                                break
+                        else:
+                            output_data[f'{predicted_col}_confidence'][sample_idx] = 0.9901  # default
```
George3d6, line 342: Setting this to 0.999 makes sense for numerical, since the confidence is given for the range, but it doesn't seem to make sense for categorical, where we want the confidence per a single categroy

George3d6 (changes requested):
Overall a few things seem wrong, but I might just be misunderstanding some logic, I'd appreicate some further on-call discussion before merging this.

PR #165, opened 2020-09-12
Sourcing historical data

George3d6:


mindsdbadmin:

mindsdb_native/libs/helpers/query_composer.py
```diff
@@ -0,0 +1,65 @@
+from mindsdb_native.libs.constants.mindsdb import *
+
+
+def create_history_query(query, tss, stats, row):
```
maximlopin, line 9: there is `test_database_history()` but I still think we need a unit test that will test this function with a few example SQL queries

mindsdb_native/libs/phases/data_extractor/data_extractor.py
```diff
@@ -88,6 +91,43 @@ def _get_prepared_input_df(self):
...
+                    df['make_predictions'] = [True] * len(df)
+
+                setup_args = deepcopy(self.transaction.lmd['setup_args'])
+
+
+                if len(df) > 1 and self.transaction.lmd['tss']['group_by'] is not None:
+                    encountered_set = set()
+                    unique_group_by_rows = []
+                    for i in range(en(df)):
+                        val_str = ''
```
maximlopin: you can use tuples instead of strings

maximlopin:
```
val_str = tuple()
for group_col in self.transaction.lmd['tss']['group_by']:
    val_str = tuple([*val_str, df.iloc[i][group_col]])
    if val_str not in encountered_set:
        encountered_set.add(val_str)
        unique_group_by_rows.append(df.iloc[i])
```

mindsdb_native/libs/phases/data_extractor/data_extractor.py
```diff
@@ -88,6 +91,43 @@ def _get_prepared_input_df(self):
...
+                        for group_col in self.transaction.lmd['tss']['group_by']:
+                            val_str += '###-------------###' + str(df.iloc[i][group_col])
+                            if val_str not in encountered_set:
+                                encountered_set.add(val_str)
+                                unique_group_by_rows.append(df.iloc[i])
+                else:
+                    unique_group_by_rows = [df.iloc[0]]
```
maximlopin, line 112: `len(df)` is equal to `0` in this `else` , so `df.iloc[0]` is an `IndexError`

mindsdb_native/libs/phases/data_extractor/data_extractor.py
```diff
@@ -88,6 +91,43 @@ def _get_prepared_input_df(self):
...
+
+                historical_df = None
+
+                for row in unique_group_by_rows:
+                    setup_args['query'] = create_history_query(setup_args['query'], self.transaction.lmd['tss'], self.transaction.lmd['stats_v2'], row)
+
+                    if historical_df is None:
```
maximlopin, line 119: this is always `True`

maximlopin (changes requested):

mindsdb_native/libs/phases/data_extractor/data_extractor.py
George3d6: Hmh, no, `historical_df` gets assigned a value in the first itteration of the loop and then the other dataframes are concated to it afterwards.

mindsdb_native/libs/phases/data_extractor/data_extractor.py
George3d6: `len(df) > 1` so length should be 1, the dataframe should never be 0

mindsdb_native/libs/helpers/query_composer.py
George3d6: Hmh, yes, it might be worthwhile unittesting this, I'll add some if I have the time tonight (should have).

mindsdb_native/libs/phases/data_extractor/data_extractor.py
George3d6: good point.

mindsdb_native/libs/phases/data_extractor/data_extractor.py
```diff
@@ -102,11 +102,11 @@ def _get_prepared_input_df(self):
                     encountered_set = set()
                     unique_group_by_rows = []
                     for i in range(en(df)):
-                        val_str = ''
+                        val_tuple = tuple()
                         for group_col in self.transaction.lmd['tss']['group_by']:
-                            val_str += '###-------------###' + str(df.iloc[i][group_col])
-                            if val_str not in encountered_set:
-                                encountered_set.add(val_str)
+                            val_tuple.append([*val_tuple,df.iloc[i][group_col]])
```
maximlopin: it should be `val_tuple = tuple([*val_tuple, df.iloc[i][group_col]])`

PR #175, opened 2020-09-16
Test for save -> load -> predict

paxcema:
This PR adds a test as general behavior check for saving and then loading a predictor, which is how a bug was discovered in the confidence estimator. 

~~I am not really sure if this is the correct place (or name), though.~~


mindsdbadmin:

requirements.txt
```diff
@@ -1,5 +1,5 @@
 psutil>=5.7.0
-lightwood >= 0.39.0
+lightwood == 0.39.0
```
George3d6: This is merging into staging so move back lightwood to previous version.

tests/unit_tests/libs/controllers/test_predictor.py
```diff
@@ -714,3 +714,18 @@ def test_user_provided_split_indices(self):
...
+                to_predict='rental_price',
+                stop_training_in_x_seconds=3,
+                advanced_args={'force_predict': True})
+
+        p = None
+        F.export_predictor(name)
+        F.import_model(f"{name}.zip")
+
+        p = Predictor(name=name)
+        predictions = p.predict(when_data={'sqft': 1000}, run_confidence_variation_analysis=True)
```
George3d6: Training a model here wastes precious time, please combine it with one of the other tests that trains a model and make a comment as to what it does.

George3d6 (changes requested):

PR #191, opened 2020-09-23
Best mixer selection

### LINKED ISSUE #131: ML model selection from native
Allow for model selection from Mindsdb, either in the Model interface phase or potentially better, in a separate `ModelSelection` phase.

The first implementation of this should look something like:

1. Train 3 x lightwood models (one with the NN mixer, one with the Boosting Mixer, one with a yet-to-be-implemented sklearn backend that just chooses between various sklearn models) 
2. Test the 3 models on the validation data and get the one with the highest accuracy (if it's close, pick the NN models since that has better confidence estimations)
3. Use that model in the future for all predictions (maybe allow the user to switch the model via the `advanced_args`

For this to be done more efficiently we should implement encoder sharing between lightwood models so that we can train them only once, but it's not needed in the first iteration.

maximlopin:
Resolves https://github.com/mindsdb/mindsdb_native/issues/131
- Iterate through all mixers and select the one that has the best accuracy on the validation data
- Encoder sharing (only one lightwood DataSource is created for all models)

tests pass on lightwood branch `lightwood/mixers-for-tags`

### DIFF
--- a/mindsdb_native/libs/backends/lightwood.py
+++ b/mindsdb_native/libs/backends/lightwood.py
@@ -313,7 +313,7 @@ class LightwoodBackend():
                 kwargs = lightwood_config['mixer']['kwargs']
 
                 kwargs['callback_on_iter'] = self.callback_on_iter
-                kwargs['eval_every_x_epochs'] = eval_every_x_epochs
+                kwargs['eval_every_x_epochs'] = eval_every_x_epochs / len(mixer_classes)
                 kwargs['stop_training_after_seconds'] = self.transaction.lmd['stop_training_in_x_seconds']
 
             self.predictor = lightwood.Predictor(lightwood_config.copy())
--- a/mindsdb_native/libs/backends/lightwood.py
+++ b/mindsdb_native/libs/backends/lightwood.py
@@ -261,7 +261,13 @@ class LightwoodBackend():
         logging.getLogger().setLevel(logging.DEBUG)
 
         predictors_and_accuracies = []
-        for mixer_class in lightwood.mixers.BaseMixer.__subclasses__():
+
+        if self.transaction.lmd['mixer_class'] is not None:
+            mixer_classes = [self.transaction.lmd['mixer_class']]
+        else:
+            mixer_classes = lightwood.mixers.BaseMixer.__subclasses__()
+
+        for mixer_class in mixer_classes:
             lightwood_config['mixer']['kwargs'] = {}
             lightwood_config['mixer']['class'] = mixer_class
 
@@ -281,7 +287,7 @@ class LightwoodBackend():
                 kwargs['eval_every_x_epochs'] = eval_every_x_epochs
                 kwargs['stop_training_after_seconds'] = self.transaction.lmd['stop_training_in_x_seconds']
 
-            self.predictor = lightwood.Predictor(lightwood_config)
+            self.predictor = lightwood.Predictor(lightwood_config.copy())
 
             self.predictor.learn(
                 from_data=lightwood_train_ds,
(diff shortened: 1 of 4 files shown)


mindsdbadmin:

mindsdb_native/libs/backends/lightwood.py
```diff
@@ -278,38 +278,95 @@ def train(self):
...
-        self.transaction.log.info('Training accuracy of: {}'.format(self.predictor.train_accuracy))
+                kwargs = lightwood_config['mixer']['kwargs']
+
+                kwargs['callback_on_iter'] = self.callback_on_iter
+                kwargs['eval_every_x_epochs'] = eval_every_x_epochs
+                kwargs['stop_training_after_seconds'] = self.transaction.lmd['stop_training_in_x_seconds']
+
+            self.predictor = lightwood.Predictor(lightwood_config.copy())
+
+            self.predictor.learn(
```
George3d6, line 321: I think this will suffer from the issue of encoders being re-trained, we should look at encoder transfer before deploying this (see issues lightwood side), unless that's already implemented and I'm wrong here ?

Alternative here might be to just:

1. Deepcopy predictor
2. Change mixer
3. Make sure lightwood keeps the old encoders which are all marked as "prepared"

mindsdb_native/libs/backends/lightwood.py
```diff
@@ -278,38 +278,95 @@ def train(self):
...
+                if eval_every_x_epochs > 200:
+                    eval_every_x_epochs = 200
+                if eval_every_x_epochs < 3:
+                    eval_every_x_epochs = 3
 
-        self.transaction.log.info('Training accuracy of: {}'.format(self.predictor.train_accuracy))
+                kwargs = lightwood_config['mixer']['kwargs']
+
+                kwargs['callback_on_iter'] = self.callback_on_iter
+                kwargs['eval_every_x_epochs'] = eval_every_x_epochs
```
George3d6: Divide by the nr of mixers.

mindsdb_native/libs/backends/lightwood.py
```diff
@@ -278,38 +278,95 @@ def train(self):
...
+                break
+        else:
+            nn_mixer_predictor, nn_mixer_predictor_accuracy = None, None
+
+        self.predictor = best_predictor
+
+        # If difference between accuracies of best predictor and NnMixer predictor
+        # is small, then use NnMixer predictor
+        if nn_mixer_predictor is not None:
+            SMALL_ACCURACY_DIFFERENCE = 0.05
```
George3d6: Make 0.01, a difference of 5% seems like it could be significant, 1% might well be noise. 

But, obviously, just a heuristic, still, our focus is on accuracy so I'd like to prioritize that for now.

mindsdb_native/libs/controllers/predictor.py
```diff
@@ -246,6 +246,7 @@ def learn(self,
                 data_split_indexes = advanced_args.get('data_split_indexes', None),
                 tags_delimiter = advanced_args.get('tags_delimiter', ','),
                 force_predict = advanced_args.get('force_predict', False),
+                mixer_class = advanced_args.get('mixer_class', None),
```
George3d6, line 249: Set this to be NNMixer + Sklearn mixer for now, since the former should be quick to train and give us a lot of potential gains on many datasets.

George3d6: Also remove `None` handling from lightwood.py and just handle it here.

George3d6 (changes requested):

mindsdb_native/libs/backends/lightwood.py
maximlopin: data source is only created once: https://github.com/mindsdb/mindsdb_native/pull/191/files/d102fdec763ac90b647fbfb7b8ce07adc9a81d4b#diff-ab4b77a7fb52eb86c1fb66a4cbe8b6abR281

so encoders are trained once in `DataSource.__init__`

mindsdb_native/libs/controllers/predictor.py
maximlopin: > Set this to be NNMixer + Sklearn mixer for now, since the former should be quick to train and give us a lot of potential gains on many datasets.

providing `mixer_class` forces usage of a single mixer

PR #200, opened 2020-09-24
Identifier detection improvements

### LINKED ISSUE #146: Identifier column based on information score
If the entropy for a given column (as provided by `compute_entropy_biased_buckets`) is low enough (e.g. < 0.01 or 0) then this column should be marked as an identifier. 

This should be doable for both categorical and numerical, but I wouldn't expect the score to get this low for numerical values.

### LINKED ISSUE #147: Identifier from AI primary key structure
We should detect the most basic type of primary key identifier, an auto incrementing key.

There's a few approaches we can try:

1. Sort a numeric int column and see if all the values are increasing by `1`.
2. Sort a numeric int column and see if all the values are increasing by a constant `n`.
3. Sort a numeric int column  and see if `x%` of the values are increasing by a constant `n` (and all the values are unique). Where we could start with `x` being rather large (e.g. 99.5) in order to avoid too many false positives.

### LINKED ISSUE #150: Figure out how to detect foreign keys
We need to get a few strategies for detecting when a column we label `identifier` is a foreign key, such that we can suggest to the user they should import data from the table[s] being referenced by them.

George3d6:
Refactored identifier detection to make it a bit easier to read, remove dead code, make it more aggressive and allow it to detect some additional type of identifier.

Also made it work on the full column, since the operations it performs are inexpensive.

Resolves #147
Resolves #150  (though maybe only partially so)
Resolves #146 (to the extent this can be done)

### DIFF
--- a/mindsdb_native/libs/helpers/text_helpers.py
+++ b/mindsdb_native/libs/helpers/text_helpers.py
@@ -170,15 +170,18 @@ def isascii(string):
     """
     return all(ord(c) < 128 for c in string)
 
+def extract_digits(point):
+    return int([char if char.isdigit() for char in str(point)])
 
 def get_pct_auto_increment(data):
     data = sorted(data)
     prev = None
     increase_by_one = 0
     for point in data:
-        if prev is not None and point is not None:
+        digits = extract_digits(point)
+        if prev is not None and digits is not None:
             try:
-                diff = int(point) - int(prev)
+                diff = int(digits) - int(digits)
                 if diff == 1:
                     increase_by_one += 1
             except Exception:
--- a/mindsdb_native/libs/helpers/text_helpers.py
+++ b/mindsdb_native/libs/helpers/text_helpers.py
@@ -148,18 +148,16 @@ def hashtext(cell):
 
 
 def _is_foreign_key_name(name):
-    for endings in ['-id', '_id', 'ID', 'Id']:
-        if name.endswith(endings):
-            return True
+    for endings in ['id', 'ID', 'Id']:
+        for add in ['-','_', ' ', '']
+            if name.endswith(add + endings):
+                return True
+    return False
 
+def _is_identifier_name(name):
     for keyword in ['account', 'uuid', 'identifier', 'user']:
         if keyword in name:
             return True
-
-    for keyword in ['id']:
-        if keyword == name:
-            return True
-
     return False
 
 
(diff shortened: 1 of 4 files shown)


mindsdbadmin:

tests/unit_tests/libs/controllers/test_predictor.py
```diff
@@ -167,8 +167,9 @@ def test_ignore_columns(self):
 
     def test_ignore_identifiers(self):
         input_dataframe = pd.DataFrame({
-            'do_use': [*range(99), 123],
+            'do_use': [*range(60), *range(40)],
             'numeric_id': list(range(100)),
+            'malicious_naming': list(range(99),200),
```
maximlopin: this will riase `TypeError`, it should be `'malicious_naming': [*range(99), 200]`

mindsdb_native/libs/helpers/text_helpers.py
```diff
@@ -170,103 +168,58 @@ def isascii(string):
...
-        #    |                        Counter({q: 7})
-        #    |                        S = entropy[7]
-        #    |                        randomness = S / np.log(1) <----- 1 unique value at this index
-        #    |
-        #    |----------------------- index 0
-        #                             Counter({w: 3, e: 4})
-        #                             S = entropy[3, 4]
-        #                             randomness = S / np.log(2) <----- 2 unique values at this index
-        #
-        # Scaling entropy by np.log(num_of_unique_values) produces a number in range [0, 1]
```
maximlopin, line 235: why is the comment deleted?

mindsdb_native/libs/helpers/text_helpers.py
```diff
@@ -170,103 +168,58 @@ def isascii(string):
...
-def get_identifier_description(data, column_name, data_subtype, other_potential_subtypes):
-    data = list(data)
+def get_pct_auto_increment(data):
+    data = sorted(data)
+    prev = None
+    increase_by_one = 0
+    for point in data:
+        if prev is not None and point is not None:
+            try:
+                diff = int(point) - int(prev)
```
maximlopin: each data point can be convertd to digit-only characters : `int([char if char.isdigit() for char in str(point)])`

this woud look something like this:
```
def to_digits(point):
    return int([char if char.isdigit() for char in str(point)])
...
if prev is not None and to_digits(point) != '':
    diff = int(to_digits(point)) - int(to_digits(prev))
```

maximlopin: this would work for data like `abcdef_1, abcdef_2, abcdef_3, ...`

mindsdb_native/libs/helpers/text_helpers.py
```diff
@@ -170,103 +168,58 @@ def isascii(string):
...
+                if diff == 1:
+                    increase_by_one += 1
+            except:
```
maximlopin: `except Exception`

mindsdb_native/libs/helpers/text_helpers.py
```diff
@@ -148,18 +148,16 @@ def hashtext(cell):
 
 
 def _is_foreign_key_name(name):
-    for endings in ['-id', '_id', 'ID', 'Id']:
-        if name.endswith(endings):
-            return True
+    for endings in ['id', 'ID', 'Id']:
+        for add in ['-','_', ' ', '']:
```
paxcema: That last `''` element is not needed, is it?

mindsdb_native/libs/helpers/text_helpers.py
George3d6: I thought this approach was no longer needed, so no point to leave the comment in. Or is there still a situation when we'd want to reintroduce this ?

mindsdb_native/libs/helpers/text_helpers.py
George3d6: Interesting take, I think it should be something else other than "auto incrementing index" though, since it's not that same thing, but we should add this approach.

mindsdb_native/libs/helpers/text_helpers.py
George3d6: Hmh, good point, the approach I have here is a bit... ahm, stupid.

George3d6: Ok, see new approach, it's closer to what I'd have wanted to achieve.

mindsdb_native/libs/helpers/text_helpers.py
maximlopin: but the code for it is still there: https://github.com/mindsdb/mindsdb_native/pull/200/files#diff-d26ce10f25a7c27ed6efdd7906661481R211

George3d6: added back in, my fault.

PR #204, opened 2020-10-05
Snowflake datasources

### LINKED ISSUE #154: Snowflake DataSource
Create a DataSource that is able to pull data from Snowflake.

In order to test this we probably want to get a tiny snowflake instantiation going on AWS and then go ahead and add the credentials for it to the secrets in github and into env variables in our local environments (+ Ignore the test when it's run locally unless those credentials are passed via env variables)

George3d6:
Added snowflake datasource, only work if installing `mindsdb_native[extra_data_sources]`

Closes #154

### DIFF
--- a/mindsdb_native/libs/data_sources/snowflake_ds.py
+++ b/mindsdb_native/libs/data_sources/snowflake_ds.py
@@ -1,0 +1,45 @@
+import os
+
+import snowflake
+
+from mindsdb_native.libs.data_types.data_source import DataSource
+
+
+class PostgresDS(DataSource):
+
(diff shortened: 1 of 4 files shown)


mindsdbadmin:

George3d6:
Connector is broken at the moment: https://stackoverflow.com/questions/63704212/could-not-build-wheels-for-snowflake-connector-python-which-use-pep-517

George3d6:
Added the DataSource as an optional dependency for now as to not break existing functionality due to the connector not working.

PR #205, opened 2020-10-05
Improve history query generation

George3d6:
When generating the historical query to timeseries, we'll only going to select rows where the order by columns are smaller than the value of the order by columns in the current row.


mindsdbadmin:

paxcema (approved):

PR #208, opened 2020-10-06
Array predictions

paxcema:
- Additional number of predictions for time series tasks

### DIFF
--- a/mindsdb_native/libs/backends/lightwood.py
+++ b/mindsdb_native/libs/backends/lightwood.py
@@ -324,9 +324,6 @@ class LightwoodBackend():
         else:
             raise Exception(f'Unknown mode specified: "{mode}"')
 
-        if self.transaction.lmd['tss']['is_timeseries']:
-            df, _ = self._create_timeseries_df(df)
-
         if self.predictor is None:
             self.predictor = lightwood.Predictor(load_from_path=self.transaction.lmd['lightwood_data']['save_path'])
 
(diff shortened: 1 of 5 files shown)


mindsdbadmin:

mindsdb_native/libs/backends/lightwood.py
```diff
@@ -328,7 +328,7 @@ def train(self):
                 self.predictor.train_accuracy
             ))
 
-            validation_predictions = self.predict('validate')
+            validation_predictions = self.predict('validate')[0]  # we only eval t+1 accuracy
```
George3d6, line 331: Hmh, this will fail, we need to add an output type check before this.

mindsdb_native/libs/backends/lightwood.py
```diff
@@ -395,37 +392,54 @@ def predict(self, mode='predict', ignore_columns=None, all_mixers=False):
...
+
+        # additional predictions for timeseries
+        timesteps = self.transaction.lmd.get('nr_predictions', 1)
+        if self.transaction.lmd['tss']['is_timeseries']:
+            for t in range(1, timesteps):
+                displaced_df = copy.deepcopy(df)
+                for target in self.predictor._output_columns:
+                    displaced_df[target] = predictions[t-1][target]['predictions']
+                # if self.transaction.lmd['tss']['use_previous_target']:
+                displaced_df, _ = self._create_timeseries_df(displaced_df)
```
George3d6, line 409: The logic here seems incorrect, the way to do it (I'd think) is something like:

1. create a new row that contains all the values from the row used to predict + the value for the `target` set to the prediction
2. append that row to `df` (or rather to a copy of it, but only make this copy onceA)
3. `displaced_df, _ = self._create_timeseries_df(copy_of_df_with_row_appended)`
4. Rinse and repeat (appending to the same df

George3d6 (changes requested):

mindsdb_native/libs/backends/lightwood.py
```diff
@@ -395,37 +392,54 @@ def predict(self, mode='predict', ignore_columns=None, all_mixers=False):
...
-            for k in model_confidence_dict:
-                model_confidence_dict[k] = [np.mean(x) for x in model_confidence_dict[k]]
-
-            for k  in model_confidence_dict:
-                formated_predictions[f'{k}_model_confidence'] = model_confidence_dict[k]
-
-            if 'confidence_range' in predictions[k]:
-                formated_predictions[f'{k}_confidence_range'] = predictions[k]['confidence_range']
+        for t in range(timesteps):
+            formated_predictions[t] = {}
```
George3d6, line 414: I don't quite understand the logic here, it might be better to:

1) have predictions == predictions[-1] for the purpose of computing the confidence (do an if timesteps > 1 and set it somewhere)
2) have formated_predictions[k] be an array of arrays [[Y1_T+1,Y1_T+2,Y1_T+3], [Y2_T+1,Y2_T+2,Y2_T+3], [Y3_T+1,Y3_T+2,Y3_T+3]] ... etc

George3d6 (changes requested):

PR #212, opened 2020-10-07
Snowflake tests

George3d6:
A test scenario for snowlfake, however can't be used atm because the env is temporary (we loss it in 30 days) and the connector is broken for python 3.8

Fixed a few datasource bugs in the process.


mindsdbadmin:

PR #237, opened 2020-10-14
Timesereis data for the UI

George3d6:
## Why

* The GUI needs to have some information for plotting a representative accuracy graph for time series problems
* The GUI needs information about the time series settings and other miscellaneous ts related stuff in the model_data

## How
* Added `timeseries` top-level key to model data (returned by `get_model_data`) which contains the timeseries settings.
* Added `test_data_plot` key containing info to plot predicted, real and the confidence range for timeseries data.
* Added `setup_args` top level key containing info about the datasource, which will include whether or not this is a live datasource in the future (not super relevant for now).

## Extra

* Made native to work with the latest lightwood staging branch, which no longer returns confidence ranges.
* Normalizing r2 regression accuracy scores between 0 and 1
* Confidence range determination via non-conformist in the model analysis phase
* Fix that makes the snowflake sdk to only be imported when the datasource is queried, since the snowflake library is very broken and importing it can result in it breaking the `request` package (yeah, weird, I too was unaware this is possible)


mindsdb_native/libs/phases/data_extractor/data_extractor.py
```diff
@@ -91,7 +91,7 @@ def _get_prepared_input_df(self):
                 # if no data frame yet, make one
                 df = self._data_from_when()
 
-            if self.transaction.lmd['setup_args'] is not None and self.transaction.lmd['tss']['is_timeseries']:
+            if self.transaction.lmd['setup_args'] is not None and self.transaction.lmd['tss']['disable_history_fetching'] and self.transaction.lmd['tss']['is_timeseries']:
```
paxcema: Shouldn't this be `and not ...['disable_history_fetching']`?

mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
```diff
@@ -46,6 +46,17 @@ def run(self):
             backend=self.transaction.model_backend
         )
 
+        # @TODO Limiting to 4 as to not kill the GUI, sample later (or maybe only select latest?)
+        if self.transaction.lmd['tss']['is_timeseries'] and len(normal_predictions[output_columns[0]]) < pow(10,4):
+            self.transaction.lmd['test_data_plot'] = {}
+            for output_column in output_columns:
+                self.transaction.lmd['test_data_plot'][col] = {
```
paxcema: Seems to me that `col` -> `output_column` would fix the failing tests

mindsdb_native/libs/phases/model_analyzer/model_analyzer.py
```diff
@@ -46,6 +46,17 @@ def run(self):
...
+                    'real': list(validation_df[output_column])
```
paxcema: Maybe we ought to return an array of real values if `nr_predictions > 1` in this key? It's redundant information that could be obtained in the GUI, though.

PR #240, opened 2020-10-15
Don't drop foreign key if it's in order_by

### LINKED ISSUE #239: Error when dropping foreign column used for order_by
It can happen in time series tasks that a column tagged to `order_by` is detected as a foreign key (imagine one with arbitrary timesteps, from `1` to `n`), and dropped in the foreign key detection step, triggering an error down the line.

paxcema:
While this fixes #239, I would like to benchmark a little more, to see whether this introduces any overfitting problems with the TS RNN encoder.

### DIFF
--- a/mindsdb_native/libs/helpers/text_helpers.py
+++ b/mindsdb_native/libs/helpers/text_helpers.py
@@ -182,7 +182,6 @@ def get_pct_auto_increment(data):
             pass
 
     int_data = sorted(int_data)
-    print(int_data)
     prev_nr = int_data[0]
     increase_by_one = 0
     for nr in int_data[1:]:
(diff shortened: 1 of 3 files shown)


George3d6 (changes requested):
I said I would do this and then I never did it.. right ? right. Sorry :(

Could you add a test for it though?

You could just make it part of another timeseries tests. Modify the order column so that it is an id and then just check:

1. Column is used for training (easy enough, implicit, it would crash if it weren't)
2. Column has identifier attributes associated with it

paxcema:
Sure thing, I'll add a test.

I also did some benchmarking and it went ok. Though we might want to add a disclaimer in the docs that it's preferrable to have real timestamps to order by.

PR #242, opened 2020-10-21
Flag for using database history

George3d6:
* Database history for timeseries problems is only used when a flag in the advanced args is specified
* Removed the `dynamic_window` argument since it's not implemented and doing so is harder than I thought.
* Removed the `use_order_col` argument since the interface is already very complex and I wanted to simplify it. We can always add it later fairly easily if needed
* Completely refactored how we reshape timeseries data to use pandas function, this should => significant speedups

## Important

There was a bug where if a `group_by` was specified or if the rows provided during `predict` were not already ordered, mindsdb would return predictions in an arbitrary order, instead of that in which the rows were given during `predict`. This basically made predicting with a group by or with unordered data return garbage. 

I added tests to catch this and the refactor of the timeseries reshape was in large part motivated by fixing this.

*Note: interface changes are described in this docs PR: https://github.com/mindsdb/mindsdb-docs/pull/115*

### DIFF
--- a/mindsdb_native/libs/backends/lightwood.py
+++ b/mindsdb_native/libs/backends/lightwood.py
@@ -21,6 +21,8 @@ class LightwoodBackend():
         self.predictor = None
         self.nr_predictions = self.transaction.lmd['tss']['nr_predictions']
         self.nn_mixer_only = False
+        self.timeseries_row_mapping = {}
+        self.timeseries_incomplete_rows = []
 
     def _create_timeseries_df(self, original_df):
         group_by = self.transaction.lmd['tss']['group_by'] if self.transaction.lmd['tss']['group_by'] is not None else []
@@ -35,7 +37,13 @@ class LightwoodBackend():
                 secondary_type_dict[col] = ColumnDataTypes.NUMERIC
 
         # Convert order_by columns to numbers
+        group_by_order_list = []
+        group_len = defaultdict(int)
         for _, row in original_df.iterrows():
+            if tuple(row[group_by]) not in group_by_order_list:
+                group_by_order_list.append(tuple(row[group_by]))
+            group_len[tuple(row[group_by])] += 1
+
             for col in order_by:
                 if row[col] is None:
                     row[col] = 0.0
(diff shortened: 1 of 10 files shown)


paxcema (approved):

mindsdb_native/libs/backends/lightwood.py
```diff
@@ -22,88 +25,88 @@ def __init__(self, transaction):
...
-        group_by = self.transaction.lmd['tss']['group_by'] if self.transaction.lmd['tss']['group_by'] is not None else []
-        order_by = self.transaction.lmd['tss']['order_by']
+    def _ts_reshape(self, original_df):
+        gb_arr = self.transaction.lmd['tss']['group_by'] if self.transaction.lmd['tss']['group_by'] is not None else []
+        ob_arr = self.transaction.lmd['tss']['order_by']
         window = self.transaction.lmd['tss']['window']
 
+        original_index_list = []
+        idx = 0
+        for _, row in original_df.iterrows():
```
paxcema: If I'm not mistaken, `df.itertuples()` would be much faster than `df.iterrows()`. Something like:

```
for row in original_df.itertuples():
    ...
```

and 

```
def _make_pred(row):
    return not hasattr(row, "make_predictions") or row.make_predictions
```

should work.

paxcema: [Source](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.iterrows.html) for `itertuples()` faster than `iterrows()` claim. [Here's](https://github.com/dimgold/Datathon-TAU/blob/master/7.6%20Pandas%20iterations%20test/Pandas_Iterations.ipynb) a very simple test, and in [this](https://stackoverflow.com/questions/24870953/does-pandas-iterrows-have-performance-issues) Q&A the graphs at the bottom seem to support this.

mindsdb_native/libs/backends/lightwood.py
```diff
@@ -22,88 +25,88 @@ def __init__(self, transaction):
...
             if self.transaction.lmd['stats_v2'][col]['typing']['data_type'] == DATA_TYPES.DATE:
                 secondary_type_dict[col] = ColumnDataTypes.DATETIME
             else:
                 secondary_type_dict[col] = ColumnDataTypes.NUMERIC
 
-        # Convert order_by columns to numbers
+        # Convert order_by columns to numbers (note, rows are references to mutable rows in `original_df`)
         for _, row in original_df.iterrows():
-            for col in order_by:
+            for col in ob_arr:
```
paxcema, line 54: As above, maybe using `itertuples()` would be faster.

mindsdb_native/libs/backends/lightwood.py
```diff
@@ -114,22 +117,34 @@ def _create_timeseries_df(self, original_df):
...
-        combined_df = pd.concat(list(ts_groups.values()))
 
         if 'make_predictions' in combined_df.columns:
             combined_df = pd.DataFrame(combined_df[combined_df['make_predictions'].astype(bool) == True])
             del combined_df['make_predictions']
 
-        return combined_df, secondary_type_dict
+        timeseries_row_mapping = {}
+        idx = 0
+        for _, row in combined_df.iterrows():
```
paxcema, line 139: As above, maybe using `itertuples()` would be faster.

tests/unit_tests/libs/controllers/test_predictor_timeseries.py
```diff
@@ -73,6 +73,9 @@ def test_timeseries(self, tmp_path):
 
         results = mdb.predict(when_data=test_file_name, use_gpu=False)
 
+        # Results should only be given for the rows with full history
+        print(results)
```
paxcema: Probably a leftover print?

tests/unit_tests/libs/controllers/test_predictor_timeseries.py
```diff
@@ -175,3 +181,57 @@ def test_keep_id_orderby(self, tmp_path):
...
+            },
+            stop_training_in_x_seconds=10,
+            use_gpu=False,
+            advanced_args={'force_predict': True}
+        )
+
+        results = mdb.predict(when_data=test_file_name, use_gpu=False)
+
+        for i, row in enumerate(results):
+            print(row['order_ai_id'], row['3_valued_group_by'], columns_test[2][i], columns_test[3][i])
```
paxcema: Leftover print?

George3d6:
I tried changing to itertuples @paxcema but it causes a bunch of issues, the biggest one being that tuples are not mutating the original dataframe, which we actually need in some cases, will change where it's doable.

paxcema:
Ahh that is actually a key point. Maybe it is not doable at all, tuples are immutable. 

Well, everything else seemed good to me.

PR #244, opened 2020-10-22
use remote database for datasource tests

maximlopin:
- use unittest in all tests
- use breakpoints for phase tests
- use remote databases in datasource tests


George3d6 (changes requested):
Switch to using the tables inside the databases for the integration tests
For the db history test keep generating a table but add a uuid to the name so there's no conflicts and drop it once the test passes

George3d6 (approved):
Good to go, @maximlopin please create issues about adding an MSSQL database and a snowflake database to the credentials and adding them to the testing suite.

As well as an issue describing the problems with the mongo tests if those are still present.

PR #249, opened 2020-10-25
Adding google cloud storage datasource

purvimisal:
Fix for mindsdb [#797](https://github.com/mindsdb/mindsdb/issues/797) 
Please let me know if this looks correct and if something should be changed about how to take credentials to access gcs bucket and file. 

I also added the extra requirements for gcs in the optional_requirements_extra_data_sources.txt . 
For testing, we will need a .json credentials file for a service account from GCS and a bucket+file on GCS that allows access for the credentials provided. 
Please let me know if anything is wrong and how I can change it. Thanks!

### DIFF
--- a/mindsdb_native/libs/data_sources/gcs_ds.py
+++ b/mindsdb_native/libs/data_sources/gcs_ds.py
@@ -1,0 +1,39 @@
+import os 
+
+from google.cloud import storage
+
+from mindsdb_native.libs.data_types.data_source import DataSource
+from mindsdb_native.libs.data_sources.file_ds import FileDS
+from mindsdb_native import F
+
+
+class GCSDS(DataSource):
+
+    def _setup(self, bucket_name, file_path, path_to_auth_json):
+
+        self._bucket_name = bucket_name
+        self._file_name = os.path.basename(file_path)
+
+        gc_client = storage.Client.from_service_account_json(json_credentials_path=path_to_auth_json)
+        bucket = gc_client.get_bucket(bucket_name)
+
+        blob = storage.Blob(file_path, bucket)
+
+        self.tmp_file_name = '.tmp_mindsdb_data_file'
+
+        with open(self.tmp_file_name, 'wb') as fw:
+            gc_client.download_blob_to_file(blob, fw)
+
+        file_ds = FileDS(self.tmp_file_name)
+        return file_ds._df, file_ds._col_map
+
+    def _cleanup(self):
+        os.remove(self.tmp_file_name)
+
+
+    def name(self):
+        return '{}: {}/{}'.format(
+            self.__class__.__name__,
+            self._bucket_name,
+            self._file_name
+        )
--- a/mindsdb_native/libs/data_sources/gcs_ds.py
+++ b/mindsdb_native/libs/data_sources/gcs_ds.py
@@ -1,6 +1,8 @@
 import os 
 
 from google.cloud import storage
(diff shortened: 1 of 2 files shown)


ZoranPandovski:
recheckcla

mindsdbadmin:

optional_requirements_extra_data_sources.txt
```diff
@@ -2,3 +2,24 @@ boto3 >= 1.9.0
...
+idna==2.10
+protobuf==3.13.0
+pyasn1==0.4.8
+pyasn1-modules==0.2.8
+pycparser==2.20
+pytz==2020.1
+requests==2.24.0
+rsa==4.6
+six==1.15.0
+urllib3==1.25.11
```
George3d6: That's *a lot* of dependencies.
Is it not possible to limit this to a smaller set of libraries that can then pull in the required extra dependencies ?
Also, can you use `>=` versions rather than `==`, it makes it easier to install mindsdb in a variety of environments.

mindsdb_native/libs/data_sources/gcs_ds.py
```diff
@@ -0,0 +1,39 @@
...
+from google.cloud import storage
+
+from mindsdb_native.libs.data_types.data_source import DataSource
+from mindsdb_native.libs.data_sources.file_ds import FileDS
+from mindsdb_native import F
+
+
+class GCSDS(DataSource):
+
+    def _setup(self, bucket_name, file_path, path_to_auth_json):
```
George3d6: Isn't there some default path to the GC credentials like there is for aws ? If so, can you default `path_to_auth_json` to that path

mindsdb_native/libs/data_sources/gcs_ds.py
```diff
@@ -0,0 +1,39 @@
...
+
+        self._bucket_name = bucket_name
+        self._file_name = os.path.basename(file_path)
+
+        gc_client = storage.Client.from_service_account_json(json_credentials_path=path_to_auth_json)
```
George3d6: Maybe allow `path_to_auth_json` to be either a  path or a json and rename it to `auth_json`, in which case you can use `json.loads(auth_json)` to read the credentials if it's a valid json and if it isn't than assume it's a file and try reading it from the file system.

George3d6 (changes requested):
Sorry for the delay in reviewing.
This looks great, made a few small notes but on the whole this is exactly what we need :)

optional_requirements_extra_data_sources.txt
purvimisal: I changed it. Sorry I had directly transferred the output from pip freeze which wasn't needed. Now I have changed it. 
google-cloud-storage and google-auth will install all the dependencies but they don't have a version as such.

mindsdb_native/libs/data_sources/gcs_ds.py
purvimisal: I haven't found any default credentials even after extensively researching so I am not sure how to go about it. The json file usually downloads in the default Download folder.

mindsdb_native/libs/data_sources/gcs_ds.py
purvimisal: I added both the options to load from a file through path as well as json itself. 
PTAL! and let me know if I should change any more.

George3d6:
Thank you @purvimisal , this looks perfect now, great work :)

Issue #148, opened 2020-09-02
Predict before analysis is done


George3d6:
Define a new quick predict functionality that should cover a few situations:

* It should directly source predictions from lightwood as soon as possible, ideally returning directly from the `ModelInterface` phase.
* It should be able to run during `ModelAnalysis` is running (while the model is still training).
* It should be able to run even if `ModelAnalysis` didn't run.

This behavior should be exposed via a new `quick_predict` advanced arg added to `predict`.

This behavior should be exposed via a `quick_predict` function added to the interface that wraps around the `predict`, passing all arguments given by the user plus force the value `True` for the `quick_predict` advanced_arg.

PR #252, opened 2020-10-27
Quick learn and quick predict

### LINKED ISSUE #149: Quick Model Training
Define a new quick learn functionality, this should work similarly to the `disable_optional_analysis` argument, but disable even more thing.

`ModelAnalysis` should be disabled fully and any part of the `DataAnalyzer` phase that's not directly needed for the `DataTransformer` phase (I think we should be able to fully disable `DataAnalyzer`).

This behavior should be exposed via a new `quick_learn` advanced arg added to `learn`.

This behavior should be exposed via a `quick_predict` function added to the interface that wraps around the `learn`, passing all arguments given by the user plus force the value `True` for the `quick_learn` advanced_arg.

George3d6:
Resolves #149 and #148 

For how to use see the tests in `quick_interface.py`


paxcema (approved):

paxcema (changes requested):
Ok, for the OpenML integration we need to expose the labels in classification tasks somehow, so we should probably restore the `['histogram']['x']` flow in the data transformer. Exposing the class names in a cleaner way (just for these `quick` methods) is also an option.

Let me know what you think @George3d6, I can do the actual change

George3d6:
@paxcema I'll expose the histograms again but the training will be delayed

PR #261, opened 2020-10-30
Conformal predictor improvements

paxcema:
- Fixed issue where the lightwood data was saved to disk twice during the ICP -> pickle transition
- Added smoothing for classification tasks (benchmarks indicate it helps)
- Restored time series support, although only showing confidence for `nr_prediction = 1`, all further timesteps only return the predicted values
- Added exception handling in `_ts_reshape` because of a Pandas bug (reported [here](https://github.com/pandas-dev/pandas/issues/37486))
- Binned significance levels in classification for better performance
- Simplified regression test
- Added classifier test

NOTE: ~~this [PR](https://github.com/mindsdb/lightwood/pull/310/) to be merged first~~ Done.


mindsdb_native/libs/backends/lightwood.py
```diff
@@ -86,7 +86,12 @@ def _ts_reshape(self, original_df):
         for i in range(len(df_arr)):
             for order_col in ob_arr + self.transaction.lmd['tss']['historical_columns']:
                 for ii in range(len(df_arr[i])):
-                    df_arr[i][order_col].iloc[ii] = [df_arr[i][order_col].iloc[ii]]
+                    try:
+                        df_arr[i][order_col].iloc[ii] = [df_arr[i][order_col].iloc[ii]]
+                    except:
```
maximlopin: `except Exception` (or the exception that pandas raises)

mindsdb_native/libs/helpers/conformal_helpers.py
```diff
@@ -4,7 +4,12 @@
 
 import pandas as pd
 import numpy as np
-import copy
+from copy import deepcopy
+
+
+def softmax(x, T=1, axis=0):
```
maximlopin: maybe softmax implemented in pytorch can be used instead

tests/unit_tests/libs/helpers/test_conformal.py
```diff
@@ -1,39 +1,69 @@
...
+    Sanity check. Confidence should be a valid probability value.
+    """
+    X, y = load_iris(return_X_y=True)
+    X_train, X_test, Y_train, Y_test = train_test_split(X, y, test_size=0.1, random_state=5)
+    target = 'medv'
+
+    x_tr = _df_from_xy(X_train, Y_train, target)
+    x_te = _df_from_xy(X_test, Y_test, target)
+
+    for label_type, type_name in zip([int, float, str], ['int', 'float', 'str']):
```
maximlopin: to get class name `__name__` property can be used:
```
>>> str.__name__
'str'
```

mindsdb_native/libs/helpers/conformal_helpers.py
paxcema: Yup, I agree. Initially opted for not adding PyTorch as a dependency, but now that we're removing Ludwig there's really no harm in doing so, right?

mindsdb_native/libs/backends/lightwood.py
paxcema: Can be more than one Exception type, I'll add the generic case

mindsdb_native/libs/helpers/conformal_helpers.py
George3d6: Pytorch install is painful, so don't add it to the requirements.txt

But it's already a dep of lightwood and will always be, so feel free to just import it without having it in the requirements.

maximlopin (approved):

Issue #223, opened 2020-10-09
Test with broken datasets


George3d6:
Run tests with datasets that we artificially break. I.e.

1. Null out values (replace with `null`, `None`, `0`)
2. Mess up a few values in a column (shifting values around columns would be an ideal way to do it)
3. Predict with data types that aren't in the training datasets (e.g. pass a string for a numeircal input)
4. Remove the target values from certain rows
5. Append random characters to dates & numbers

... etc

Initially we just need to validate this doesn't break stuff, than we can start looking into the behavior exhibited (i.e. if mindsdb is still able to properly salvage the datasets by ignoring the corrupt values)

PR #265, opened 2020-11-02
Test with broken datasets

maximlopin:
https://github.com/mindsdb/mindsdb_native/issues/223

- add `break_dataset` function and use it in datasource tests

### DIFF
--- a/mindsdb_native/libs/data_types/data_source.py
+++ b/mindsdb_native/libs/data_types/data_source.py
@@ -31,10 +31,6 @@ class DataSource:
     def _cleanup(self):
         pass
 
-    @property
-    def df(self):
-        return self._df
-
     def set_subtypes(self, data_subtypes):
         """
         :param data_subtypes: dict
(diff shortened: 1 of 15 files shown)


mindsdb_native/libs/data_types/data_source.py
```diff
@@ -31,10 +31,6 @@ def _setup(self, df, **kwargs):
     def _cleanup(self):
         pass
 
-    @property
-    def df(self):
-        return self._df
-
```
StpMax, line 37: mindsb use it in one place: https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/interfaces/datastore/datastore.py#L179

How we should do: restore property in native, or change to ds._df in mindsdb?

maximlopin: my bad, I have only checked references to it in native, I will restore the property

StpMax (approved):

Issue #283, opened 2020-11-06
Cant train predictor with multiple targets from CH DS


StpMax:
When i try predict multiple targets using CH DS, i get error:
```
Traceback (most recent call last):
  File "<string>", line 5, in <module>
  File "/home/maxs/dev/mdb/venv_new/sources/mindsdb_native/mindsdb_native/libs/controllers/predictor.py", line 304, in learn
    self.transaction.run()
  File "/home/maxs/dev/mdb/venv_new/sources/mindsdb_native/mindsdb_native/libs/controllers/transaction.py", line 239, in run
    self._run()
  File "/home/maxs/dev/mdb/venv_new/sources/mindsdb_native/mindsdb_native/libs/controllers/transaction.py", line 235, in _run
    raise e
  File "/home/maxs/dev/mdb/venv_new/sources/mindsdb_native/mindsdb_native/libs/controllers/transaction.py", line 198, in _run
    self._call_phase_module(module_name='DataCleaner')
  File "/home/maxs/dev/mdb/venv_new/sources/mindsdb_native/mindsdb_native/libs/controllers/transaction.py", line 165, in _call_phase_module
    ret = module(self.session, self)(**kwargs)
  File "/home/maxs/dev/mdb/venv_new/sources/mindsdb_native/mindsdb_native/libs/phases/base_module.py", line 53, in __call__
    ret = self.run(**kwargs)
  File "/home/maxs/dev/mdb/venv_new/sources/mindsdb_native/mindsdb_native/libs/phases/data_cleaner/data_cleaner.py", line 71, in run
    self._remove_missing_targets(df)
  File "/home/maxs/dev/mdb/venv_new/sources/mindsdb_native/mindsdb_native/libs/phases/data_cleaner/data_cleaner.py", line 22, in _remove_missing_targets
    df.dropna(subset=self.transaction.lmd['predict_columns'], inplace=True)
  File "/home/maxs/dev/mdb/venv_new/lib/python3.6/site-packages/pandas/core/frame.py", line 4994, in dropna
    raise KeyError(list(np.compress(check, subset)))
KeyError: ['location']
```
I know we want to remove multiple targets prediction, but we dont do it yet, ant this is pretty strange getting this error only with CH datasource. To reproduce it, need have connection to CH with test_data.home_rentals ds:
```
from mindsdb_native import ClickhouseDS
ds = ClickhouseDS(
    query='select * from test_data.home_rentals limit 100',
    user='default',
    password='',
[...]

PR #285, opened 2020-11-06
exclude target columns from columns_to_ignore

maximlopin:
Should resolve https://github.com/mindsdb/mindsdb_native/issues/283

### DIFF
--- a/mindsdb_native/libs/phases/data_cleaner/data_cleaner.py
+++ b/mindsdb_native/libs/phases/data_cleaner/data_cleaner.py
@@ -5,17 +5,13 @@ from mindsdb_native.libs.constants.mindsdb import TRANSACTION_LEARN
 
 
 class DataCleaner(BaseModule):
-    def _get_useless_columns(self, df):
+    def _get_empty_columns(self, df):
         empty_columns = []
-        zero_information_columns = []
         for col_name in df.columns.values:
             if len(df[col_name].dropna()) < 1:
                 empty_columns.append(col_name)
                 self.log.warning(f'Column "{col_name}" is empty ! We\'ll go ahead and ignore it, please make sure you gave mindsdb the correct data.')
-            elif len(set(df[col_name])) == 1:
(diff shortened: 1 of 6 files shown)


George3d6 (changes requested):
Can you add a test for this, something like:

```
from mindsdb_native import ClickhouseDS
import mindsdb_native

ds = ClickhouseDS(
    query='select * from test_data.home_rentals limit 20',
   our_credentials
)
mindsdb_native.Predictor(name='zzz').learn(
    from_data=ds,
    to_predict=['rental_price', 'location'],
    stop_training_in_x_seconds=3,
    use_gpu=False
)
```

Ought to do the trick I think

mindsdb_native/libs/phases/data_cleaner/data_cleaner.py
```diff
@@ -58,6 +58,13 @@ def run(self):
         self.transaction.lmd['zero_information_columns'] = zero_information_columns
         self.transaction.lmd['columns_to_ignore'].extend(empty_columns)
 
+        # Exclude target columns
+        self.transaction.lmd['columns_to_ignore'] = list(set(
+            self.transaction.lmd['columns_to_ignore']
+        ).difference(
+            self.transaction.lmd.get('predict_columns', [])
+        ))
```
George3d6: Would it not be better to instead *not* run the `unique identifier` analysis for the target columns instead ? Since in Max's bug, I assume that's where the issue is happening.

George3d6 (changes requested):

mindsdb_native/libs/phases/data_cleaner/data_cleaner.py
maximlopin: it's not ran for `force_columns_usage[]` columns and for `predict_columns[]`
I decided to remove the `zero_information_columns` key because it kind of repeats the unique identifier detection logic

PR #293, opened 2020-11-10
Normalized confidence estimation through selfaware network

paxcema:
- Extends the confidence estimation pipeline to use Lightwood's selfaware neural network (when available) as normalizer for the nonconformity scores. In practice, this translates into dynamic interval widths for each prediction.
- Fixes a small bug in `use_mixers` advanced argument handling


mindsdb_native/libs/helpers/conformal_helpers.py
```diff
@@ -120,3 +121,32 @@ class probability estimates
         except KeyError:
             # Not all mixers return class probabilities yet
             return ys
+
+
+class SelfawareNormalizer(BaseScorer):
+    def __init__(self, model, fit_params=None):
+        super(SelfawareNormalizer, self).__init__()
+        self.model = model
```
George3d6: Could lead to some issues saving this (i.e. pickling and reloading)

mindsdb_native/libs/helpers/conformal_helpers.py
```diff
@@ -120,3 +121,32 @@ class probability estimates
...
+
+    def fit(self, x, y):
+        """No fitting is needed, as we instantiate this object
+        once the self-aware NN is already trained in Lightwood."""
+        pass
+
+    def score(self, true_input, y=None):
+        df = _df_from_x(true_input, self.columns)
+        self.model.config['include_extra_data'] = True
+        model_output = self.model.predict(df)
```
George3d6: Doesn't this lead to extra pointless predicting? As in, a prediction being called once in here when the prediction was already made externally? (Also, in hindsight, is this an issue with the nc object as well? I.e. are we running 3x predictions every time we predict now?)

paxcema: Yes, it does. A reformat that reuses predictions would result in _much_ simpler logic. Would you agree that's for a separate PR? Or do you prefer I make those changes in here?

George3d6: Please do them here, otherwise it increases the time for this to run more and more :<

Issue #294, opened 2020-11-11
Database datasource should *always* require a query


George3d6:
Currently, you can pass a `table` argument to a database datasource instead of a `query` and it will construct a query for you... this proves to be more annoying to factor in than it's worth and nobody is using this. So, for all SQL datasources, we should:

1. Make `query` the first argument and have it be mandatory
2. Remove the `table` argument
3. Remove all logic that deals with the `query is None` situations.

PR #295, opened 2020-11-11
Lazy and filter-able

### LINKED ISSUE #96: Add filter capabilities to the DataSource
`Warning: This issue is intended as a test to be given to candidates for a job at Mindsdb, please do not start working on it, unless we've message you telling you to do so.`

Mindsdb has a bunch of DataSources: https://github.com/mindsdb/mindsdb_native/tree/stable/mindsdb_native/libs/data_sources
All inheriting from the main `DataSource` class: https://github.com/mindsdb/mindsdb_native/blob/stable/mindsdb_native/libs/data_types/data_source.py

We need to add a `filter` function to all of these data sources, this should be able to take a list containing tuples of `[column_name, operator, value]` and filter (using the `WHERE` clause) based on those tuples. It should also accept a second argument (limit) which will limit the number of rows being returned by the `filter` method.

The filtered rows should be returned as part of a pandas data frame.

For a lot of datasources, this can be a very simple method using the internal pandas data-frame, but for databases-based datasource (mariadb, postgres, mysql, clickhouse) it might be wiser to "alter" the internal SQL statement used to query the database in order to add the filters & limit there and get the data more efficiently that way (with fallback to filtering the data frame).

You need not provide a query-based implementation for every single database-based datasource, it's sufficient to do so for one or two of them if the method is conceptually sound and could apply to every single one of them.

Make sure all datasource (even the database based one) can implement `filter` on the internal pandas dataframe if you haven't implemented it using database queries on that particular datasource.

Examples of usage:

`filter([('index','>',800),('name','=','John Doe')])`

`filter(where=[('foo','LIKE','%bar%')], limit=250)`

### LINKED ISSUE #276: Lazy datasources
Datasources should be made lazy, such that the data is only read when it's needed. Fairly easy in principle, in practice we need to make sure it's well tested since a bug here would affect all of mindsdb.

Make sure that there's good test coverage on this one.

Some of the coding-challenge submissions included solutions for this, might use them here.

George3d6:
* Datasource are now "lazy", in that the internal data frame is not initialized until the `.df` property or a property that depends on it is called
* Datasources now have a test method, which works on the underlying pandas data frame for nonsql datasources and by altering the internal query for sql datasources
 
Tests:
* Filtering tests on clickhouse DS for the SQL case
* Filtering tests on the mongo DS for the non-SQL case

Some other slight refactorings done in order to achieve this.

Closes #96 
Closes #276 

Mish-mash from the coding challenges I received from 3 people (Illya, chico and antonio) but extended the functionality described in #96 to achieve #276 properly and cleaned up some bugs.

Followup issue: https://github.com/mindsdb/mindsdb_native/issues/294


mindsdb_native/libs/data_types/data_source.py
```diff
@@ -66,7 +108,71 @@ def drop_columns(self, column_list):
...
+
+            if 'query' in kwargs:
+                kwargs['query'] = query
+            else:
+                args[0] = query
+
+            if get_col_map:
+                return self._setup(*args, **kwargs)
+            else:
+                return self._setup(*args, **kwargs)[0]
```
StpMax: I'm afraid this will only work in simple cases, because:

1. you can parse only simple select:
```
p.parse("SELECT row_number() OVER (ORDER BY id) as nom FROM TABLE")
error
```

2. back query assemble possible not in any case:
```
p.format(p.parse(" with a as (select b from c) select * from a"))
'SELECT * FROM a'
```

3. Parser dont know about syntax features of databases. In mysql insead 'LIMIT x OFFSET y' user can use OFFSET x, y:
```
p.parse("select a from b limit 5, 5")
error
```

4. You will loose escaping on back assemble:
```
p.format(p.parse("select a from `b b` "))
'SELECT a FROM "b b"'
```

after it, you use `query = query.replace('"',"'")`, so we get transform:
```
select a from `b b`  ->  'SELECT a FROM 'b b'
```

StpMax (changes requested):

mindsdb_native/libs/data_types/data_source.py
George3d6: Most queries will be simple cases, but yeah, we should have a fallback where if the query fails we filter on the data frame.

As for the escaping, that is an issue I have to deal with better :(

George3d6: Ok, I added fallback to  data frame filtering when the query fails to parse or to execute, and I added 2x new test scenarios to validate this.

In the process of doing this I also made some adjustments to data frame filtering.

PR #301, opened 2020-11-16
Test with broken datasets

maximlopin:
https://github.com/mindsdb/mindsdb_native/issues/223


George3d6 (approved):
Maybe specify the sampling arguments rather than doing the `limit=500`, because the limit will always select the first 500 rows, while sampling will try to sample in an evenly distributed fashion? I can't figure out how if at all useful this would be though.

PR #303, opened 2020-11-17
Improvement to what will no longer be telemetry

George3d6:


mindsdb_native/libs/helpers/general_helpers.py
```diff
@@ -40,13 +41,41 @@ def default(self, obj):
         else:
             return super().default(obj)
 
+def _get_mindsdb_stauts(run_env):
```
ilia-tsyplenkov: Is it typo `stauts`? If yes, please fix it

ilia-tsyplenkov (changes requested):

Issue #298, opened 2020-11-13
Low accuracy for concrete_sctength dataset


StpMax:
Last staging of all, no gpu

I get accuracy = 1% for column `concrete_strength` in `concrete_strength` dataset:
```
from mindsdb_native import FileDS, Predictor, F
ds = FileDS('benchmarks/datasets/concrete_strength/data.csv')
p = Predictor(name='cs')
p.learn(from_data=ds, to_predict=['concrete_strength'])
F.get_model_data('cs')['accuracy']
```

paxcema:
Predictions look okay, so I guess the issue here is that the logic behind the reported 1% accuracy is faulty.

maximlopin:
> Predictions look okay, so I guess the issue here is that the logic behind the reported 1% accuracy is faulty.

`ProbabilisticValidator` returns 0.01 accuracy, i'm trying to figure out why

paxcema:
I ran some tests and it's the `AccStats.get_accuracy_stats()` method that is faulty. As it stands, I'd probably opt for reporting the accuracy directly from [`normal_accuracy`](https://github.com/mindsdb/mindsdb_native/blob/aa9346a340f4c1bc7641b0919f679f0e2c4278ef/mindsdb_native/libs/phases/model_analyzer/model_analyzer.py#L41).

maximlopin:
With this (https://github.com/mindsdb/mindsdb_native/pull/321) fix the accuracy is ~0.3 but still seems wrong.

For some reason `confidence_range[0]` is sometimes greater than `confidence_range[1]`

George3d6:
See https://github.com/mindsdb/mindsdb_native/issues/339 as a followup.

Issue #259, opened 2020-10-29
Remove Probabilistic Validator and use nonconformist


George3d6:
Remove the probabilistic validator and replace it in all cases with the nonconformist model. If that can't be used for certain edge-case scenarios (e.g. images in/out, rich-text out) then just forgo having confidence in those for now.

PR #313, opened 2020-11-20
Remove ProbabilisticValidator

paxcema:
Related: #259 

- Removes ProbabilisticValidator, confidence is now entirely estimated with conformal predictors. 
- Tasks with categorical tags unsupported as of now. 
- Repackaged remaining accuracy methods into AccStats class.
- Pending: use nonconformist confidence ranges instead of Lightwood's for accuracy stats. This should also solve #298.

### DIFF
--- a/mindsdb_native/libs/data_types/transaction_output_row.py
+++ b/mindsdb_native/libs/data_types/transaction_output_row.py
@@ -44,18 +44,20 @@ class TransactionOutputRow:
                 answers[pred_col]['all_predicted_values'] = prediction_row[pred_col]
             else:
                 answers[pred_col]['predicted_value'] = prediction_row[pred_col]
-    
-            answers[pred_col]['confidence'] = round(prediction_row[f'{pred_col}_confidence'], 4)
 
-            quality = 'very confident'
-            if answers[pred_col]['confidence'] < 0.8:
-                quality = 'confident'
-            if answers[pred_col]['confidence'] < 0.6:
-                quality = 'somewhat confident'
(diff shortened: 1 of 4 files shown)


mindsdb_native/libs/data_types/transaction_output_row.py
```diff
@@ -45,23 +45,19 @@ def explain(self):
             else:
                 answers[pred_col]['predicted_value'] = prediction_row[pred_col]
 
-
-            if f'{pred_col}_model_confidence' in prediction_row:
-                answers[pred_col]['confidence'] = (prediction_row[f'{pred_col}_model_confidence'] * 3 + prediction_row[f'{pred_col}_confidence'] * 1)/4
```
George3d6, line 50: This will affect confidence quite heavily, since no we are weighting just nonconformist whereas before, at least judging by this, the non-conformist confidence weight was just 1/4, the rest being the selfaware weight (model_confidence)

Or am I wrong here ?

George3d6: I may well be

paxcema: No, you are right it affects heavily. I did some benchmarks which indicate it's an improvement (7.7% over 13 datasets), but we could revert it back for the time being and benchmark further (disclaimer: I only tested numerical tasks).

paxcema: Did some more testing, it definitely seems like an improvement. Mean 11.49% increase in the confidence accuracy across 19 datasets.

Issue #139, opened 2020-08-27
Limit of columns for ignorable input analysis


paxcema:
When a dataset has a big amount of columns, like the [micro_mass](https://www.openml.org/d/1515) or [bioresponse](https://www.openml.org/d/4134) datasets, the model analyzer phase takes quite a long to compute. We should define a limit on the number of columns over which we automatically skip this step, unless overridden by the user through an advanced argument.

PR #315, opened 2020-11-23
Skip model analyzer if too many columns

maximlopin:
Closes https://github.com/mindsdb/mindsdb_native/issues/139
Closes https://github.com/mindsdb/mindsdb_native/issues/338

- Quick learn is enabled when data has too many columns
- Remove `quick_predict` method. Now, if you trained predictor with `Predictor.quick_learn`, `Predictor.predict` will have `advanced_args['quick_predict']` set to `True`.
- `evaluate_regression_accuracy` now only returns `r2_score`
- Fix output of `Predictor.predict` for models trained with `quck_learn`

### DIFF
--- a/mindsdb_native/libs/controllers/transaction.py
+++ b/mindsdb_native/libs/controllers/transaction.py
@@ -234,7 +234,7 @@ class LearnTransaction(Transaction):
             # quick_learn can be set to False explicitly
             if self.lmd['quick_learn'] is None:
                 n_cols = len(self.input_data.columns)
-                n_cells = n_cols * len(self.input_data._sample_df)
+                n_cells = n_cols * len(self.input_data._sample_df or self.input_data.data_frame)
                 if n_cols >= 80 and n_cells > int(1e5):
                     self.log.warning('Data has too many columns, setting quick_learn to True')
                     self.lmd['quick_learn'] = True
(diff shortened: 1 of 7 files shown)


mindsdb_native/libs/controllers/transaction.py
```diff
@@ -223,10 +223,13 @@ def _run(self):
             self.save_metadata()
             self._call_phase_module(module_name='ModelInterface', mode='train')
 
+            MODEL_ANALYZER_COLUMNS_LIMIT = 100
```
paxcema: For now this will do, but I feel like it might be too high a limit. We should find more datasets on the bigger side and see what's the best threshold.

George3d6 (changes requested):
Please add a test for this @maximlopin , currently I'd expect `predict` calls to fail if model analysis is skipped.

Also, we can still run "some" parts of the model analysis, we just don't need to run the bits where we ignore arbitrary columns.

On the whole though, I'd say leave this as is and wait for the old p validator to be removed before continuing.

Or merge *if* you add a test for this that includes a `predict` call and it passes, but again, 90% sure that if this happens `predict` will fail.

George3d6 (changes requested):
This will currently fail.

In order for this to work you need to make the normal `.predict` call work with models trained with `.quick_learn`... there should actually be an issue about this.

Also, add tests for this, that way this failure would have been caught :)

George3d6 (changes requested):
1. Bring back quick_predict for reasons discussed in slack (basically it's important to have the method since it might include speedups at predict time that can't be achieved with the normal predict due to the amount of abstraction we use there)

PR #317, opened 2020-11-23
Positive numerical tasks

paxcema:
Avoid negative predictions for positive numerical targets by:
- Detecting positive numerical domains based on histogram
- Sending corresponding flag to Lightwood numerical encoder (see [this PR](https://github.com/mindsdb/lightwood/pull/318))

The PR also adds a small test to detect if this is working as intended.

### DIFF
--- a/mindsdb_native/libs/controllers/transaction.py
+++ b/mindsdb_native/libs/controllers/transaction.py
@@ -379,7 +379,12 @@ class PredictTransaction(Transaction):
                                 diff = sample[1, idx] - sample[0, idx]
                                 if diff <= tolerance:
                                     output_data[f'{predicted_col}_confidence'][sample_idx] = significance
-                                    output_data[f'{predicted_col}_confidence_range'][sample_idx] = list(sample[:, idx])
+                                    conf_range = list(sample[:, idx])
+
+                                    # for positive numerical domains
+                                    if self.lmd['stats_v2'][predicted_col].get('positive_domain', False):
+                                        conf_range[0] = max(0, conf_range[0])
+                                    output_data[f'{predicted_col}_confidence_range'][sample_idx] = conf_range
                                     break
                             else:
                                 output_data[f'{predicted_col}_confidence'][sample_idx] = 0.9901  # default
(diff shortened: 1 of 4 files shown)


mindsdb_native/libs/phases/data_analyzer/data_analyzer.py
```diff
@@ -299,18 +299,22 @@ def run(self, input_data):
...
+                            outlier_values=outliers,
+                            hist_x=histogram['x'],
+                            hist_y=histogram['y'],
+                            percentage_buckets=percentage_buckets,
+                            col_stats=stats_v2[col_name]
+                        ),
+                        'description': """Potential outliers can be thought as the "extremes", i.e., data points that are far from the center of mass (mean/median/interquartile range) of the data."""
+                    }
+                    # specify positive numerical domain
+                    if stats_v2[col_name]['histogram']['x'][0] >= 0:
```
maximlopin, line 315: is it assumed that `histogram['x']` is sorted?

paxcema: Yes. This is only for the numerical data type, and from what I can gather, `np.histogram` returns sorted bins.

mindsdb_native/libs/controllers/transaction.py
```diff
@@ -359,7 +359,12 @@ def run(self):
                                 diff = sample[1, idx] - sample[0, idx]
                                 if diff <= tolerance:
                                     output_data[f'{predicted_col}_confidence'][sample_idx] = significance
-                                    output_data[f'{predicted_col}_confidence_range'][sample_idx] = list(sample[:, idx])
+                                    conf_range = list(sample[:, idx])
+
+                                    # for positive numerical domains
+                                    if self.lmd['stats_v2'][predicted_col].get('positive_domain', False):
+                                        conf_range[0] = max(0, conf_range[0])
```
maximlopin, line 366: what if `conf_range` at indices greater than `0` has negative values?

paxcema: ~~Good catch, that _could_ happen~~

paxcema: Scratch that, it cannot happen. We get a guarantee from Lightwood that predictions are `>= 0` for positive domains, and the sigma from the conformal predictor can never be negative. Still, I'll add a test for this.

maximlopin (approved):

Issue #319, opened 2020-11-25
data_source key for preidctors


George3d6:
`data_source` key is a bit of a misnomer, it's more like a "data source type plus sourcing information key"... what should be contained in this key is:

1. The type of the datasource
2. The "principal" argument that helped us source the data

More specifically, 1) will just be a "user friendly" version of the datasource names:
* File
* Clickhouse
* Pandas
* Snowflake
* Mysql
* ... etc

2) will be the filename or URL with the csv for `FileDs` and the `query` argument for database datasources.

If 2) can't be sourced (e.g. if no `query` argument was used) just ignore it.

Examples:

`'data_source': 'File - /home/george/hdi.csv'`
...
`'data_source': 'Clickhouse - SELECT * FROM diamond_pricing WHERE location="Stockholm"'`
...
`'data_source': 'Postgres'` (In this case the value of `query` couldn't be found)
...
`'data_source': 'Pandas'` (In this case we don't have a 2nd value for Dataframe datasource)

..............

This `data_source` key should be made available as a top level key for the predictor both in the `get_model_data` and `get_models` functions.

.................

Once this is implemented please try training a few predictors with Scout and going to the `PREDICTORS` tab in Scout and seeing if an expected value is displayed under the "Datasource" column for each predictor, since this change is mainly intended to help visualization.

PR #320, opened 2020-11-25
Change ds names

maximlopin:
closes https://github.com/mindsdb/mindsdb_native/issues/319
- change datasource names
- `get_model_data` and `get_models` now return datasource names as described in the issue


mindsdb_native/libs/helpers/multi_data_source.py
```diff
@@ -12,16 +14,11 @@ def getDS(from_data):
...
-        from_ds = FileDS(from_data)
-        if from_ds is None:
-            log.error('No data matched the input data')
-
-    return from_ds
+        return DataSource(from_data)
+    elif isinstance(from_data, str):
+        if os.path.isfile(from_data) or from_data.startswith('http:') or from_data.startswith('https:'):
+            return FileDS(from_data)
+    raise ValueError('from_data must be one of: [DataSource, DataFrame, file path]')
```
George3d6: also `url` is also valid option

mindsdb_native/libs/data_sources/s3_ds.py
```diff
@@ -39,8 +39,4 @@ def query(self, q):
         return file_ds.df, file_ds._col_map
 
     def name(self):
-        return '{}: {}/{}'.format(
-            self.__class__.__name__,
-            self.bucket_name,
-            self.file_path
-        )
+        return 'S3 - {}/{}'.format(self.bucket_name, self.file_path)
```
George3d6: should be bucket_name + file path

mindsdb_native/libs/data_sources/file_ds.py
```diff
@@ -203,4 +202,5 @@ def _getDataIo(self, file):
             # No file type identified
             return data, None, dialect
 
-   
\ No newline at end of file
+    def name(self):
+        return 'File, {}'.format(self.file)
```
George3d6: One small change needed here, only use the last node in self.file.

So, if self.file is `/home/george/hdi.csv` make it only `hdi.csv` ... different from what I originally  said but I just realized we'll have problems with exposing internal pathing in mindsdb server otherwise and this will confuse people.

George3d6 (changes requested):

PR #321, opened 2020-11-26
AccStats low accuracy for concrete_strength dataset

maximlopin:
https://github.com/mindsdb/mindsdb_native/issues/298


George3d6 (commented):
I don't quite understand what this is doing?

Is it computing an accuracy that's averaged per-bucket rather than per nr of examples?

How does it help with the accuracy for this dataset?

PR #324, opened 2020-11-27
fix timeseries

maximlopin:
Attention to these lines of code:

**Don't set values to a data frame copy** (I had pandas warnings on Windows 10).
https://github.com/mindsdb/mindsdb_native/pull/324/files#diff-d2bbc408388eb96dfb5b29340f059abb66cbfe305bb99616c4a9ff7eed03668aR78

**If using row label and** `df.at` **always works why not use it without the** `if else`.
https://github.com/mindsdb/mindsdb_native/pull/324/files#diff-d2bbc408388eb96dfb5b29340f059abb66cbfe305bb99616c4a9ff7eed03668aR84

### DIFF
--- a/mindsdb_native/libs/phases/model_interface/lightwood_backend.py
+++ b/mindsdb_native/libs/phases/model_interface/lightwood_backend.py
@@ -72,10 +72,12 @@ class LightwoodBackend:
         else:
             df_arr = [original_df]
 
+        df_arr = [dataframe.astype(object) for dataframe in df_arr]
+
         # Make type `object` so that dataframe cells can be python lists
         for i in range(len(df_arr)):
             for hist_col in ob_arr + self.transaction.lmd['tss']['historical_columns']:
-                df_arr[i][hist_col] = df_arr[i][hist_col].astype(object)
+                df_arr[i].loc[hist_col] = df_arr[i][hist_col].astype(object)
 
         # Make all order column cells lists
         for i in range(len(df_arr)):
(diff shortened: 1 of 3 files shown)


George3d6:
> dataframe should be converted to astype(object) before setting sequences to its cells

Why ?

> DataFrame.__setitem__ was used instead of DataFrame.loc

Is there a difference ?

maximlopin:
> dataframe should be converted to astype(object) before setting sequences to its cells

because otherwise pandas will give an error like "can't set value to a sequence..."

> DataFrame.setitem was used instead of DataFrame.loc

no difference but pandas will not give a warning that you're setting a value to a copy of dataframe, the value will be set to the original dataframe

mindsdb_native/libs/phases/model_interface/lightwood_backend.py
```diff
@@ -75,18 +75,15 @@ def _ts_reshape(self, original_df):
...
                 for ii in range(len(df_arr[i])):
-                    try:
-                        df_arr[i][order_col].iloc[ii] = [df_arr[i][order_col].iloc[ii]]
-                    except Exception:
-                        # Needed because of a pandas bug that causes above to fail for small dataframes
-                        label = df_arr[i].index.values[ii]
-                        df_arr[i].at[label, order_col] = [df_arr[i].at[label, order_col]]
+                    # Needed because of a pandas bug that causes above to fail for small dataframes
+                    label = df_arr[i].index.values[ii]
+                    df_arr[i].at[label, order_col] = [df_arr[i].at[label, order_col]]
```
George3d6, line 86: I'm awfully afraid the try/except was there for performance reasons, but for now let's change it since we'll have to revamp this bit anyway as we are moving towards optimizing the ts reshape even further.

George3d6 (approved):

PR #325, opened 2020-11-27
"Apply function to column" advanced argument

maximlopin:
in **arrivals** (https://github.com/mindsdb/private-benchmarks/tree/master/benchmarks/datasets/arrivals) dataset,

```
1982-1
1982-2
1982-10
```

would be sorted to
```
1982-1
1982-10
1982-2
```

with "apply_to_columns" advanced argument you can define how you want to treat this column is it will be sorted the right way:
```
1982-1
1982-2
1982-10
```


maximlopin:
this function can now be passed to apply to column `'T'`:
```
def f(date):
    year, month = date.split('-')
    return datetime(year=int(year), month=int(month), day=1).timestamp()
```

George3d6:
So the issue here is that we are ordering by this column but it gets detected as "Categorical" ?

If so, even though this advanced arg might be useful, I'd rather handle the issue by making sure it correctly gets identified as a date.

Issue #336, opened 2020-12-02
Take out language analysis (temporarily)


George3d6:

torrmal:
we should remove the flair dependency too

George3d6:
Apparently not possible since language detection is still used as the final arbiter for whether or not a column *is* text and what type of text.

For now I've merged it like this since it will remove some useless computation, I'll leave this issue open since it might be possible to get rid of it entirely.

PR #349, opened 2020-12-11
Take out language analysis

maximlopin:
Closes https://github.com/mindsdb/mindsdb_native/issues/336

- Don't compute language stats for text data type

The following keys are removed from stats_v2[col_name]
- `'avg_words_per_sentence'`
- `'word_dist'`
- `'nr_words_dist'`
- `'lang_dist'`


George3d6 (changes requested):
Please remove the import + dependency in requirements.txt on flair and langdetect

PR #357, opened 2020-12-16
turn on/off telemetry endpoint

ilia-tsyplenkov:
Closes #993

- **What**
   - Need ability to turn telemetry (like check for updates) on/off via API
- **How**
   - Created a separate endpoint in [mindsdb](https://github.com/mindsdb/mindsdb/pull/1011)
   - Telemetry moves to disabled state when telemetry.lock file is created in storage dir. It is enabled otherwise
   - It looks a little bit strage, that needs to create smth to disable telemetry. But otherwise (when needs to create lock file on filesystem to enable it) CHECK_FOR_UPDATE=1 value doesn't make any affect by default.

### DIFF
--- a/mindsdb_native/__init__.py
+++ b/mindsdb_native/__init__.py
@@ -77,7 +77,7 @@ MindsDB = Predictor
 try:
     from mindsdb_native.libs.helpers.general_helpers import check_for_updates
     from mindsdb_native.config import CONFIG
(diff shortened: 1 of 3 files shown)


StpMax:
here need add check too:
https://github.com/mindsdb/mindsdb_native/blob/stable/mindsdb_native/__init__.py#L80

StpMax (approved):

PR #358, opened 2020-12-17
Issue 347: wrong dialect detection

### LINKED ISSUE #347: Wrong CSV parsing
Last staging of all.

subset of CSV:
```
PronounceGIF
"With a hard ""g,"" like ""gift"""
```
On upload this CSV i get error here: https://github.com/mindsdb/mindsdb_native/blob/stable/mindsdb_native/libs/data_sources/file_ds.py#L79
reason is `len(file_list_data) != len(col_map)`
```
len(file_list_data[1]) == 2  # ['"With a hard ""g', '"" like ""gift"""']
len(col_map) = 1
```
Full CSV:
[Cleaned.zip](https://github.com/mindsdb/mindsdb_native/files/5666230/Cleaned.zip)

ilia-tsyplenkov:
Closes #347 
- **What**
    - there is a data parsing error if column contains data in format:

             ```
            PronounceGIF
            "With a hard ""g,"" like ""gift"""
            ```

- **Why**
   - Root cause is that csv.Sniffer that used to deduce data dialect works with headers row (not data row) as template

- **How**
   - data row provided to csv.Sniffer as template if possible

### DIFF
--- a/mindsdb_native/libs/data_sources/file_ds.py
+++ b/mindsdb_native/libs/data_sources/file_ds.py
@@ -189,20 +189,17 @@ class FileDS(DataSource):
             first_few_lines = []
             i = 0
 
-            # need to have some templates to deduce a dialect
+            # need to have sample to deduce a dialect
             # but it is not a good idea to deduce dialect by header row
             # data[0]
-            for line in data:
-                if line in ['\r\n','\n']:
-                    continue
+            for i, line in enumerate(data):
                 first_few_lines.append(line)
-                i += 1
                 if i > 10:
                     break
 
             accepted_delimiters = [',','\t', ';']
 
-            #provide template from data if it is possible
+            #provide sample from data if it is possible
             dialect = csv.Sniffer().sniff(''.join(first_few_lines), delimiters=accepted_delimiters)
             data.seek(0)
             # if csv dialect identified then return csv
--- a/mindsdb_native/libs/data_sources/file_ds.py
+++ b/mindsdb_native/libs/data_sources/file_ds.py
@@ -162,7 +162,7 @@ class FileDS(DataSource):
             else:
                 data = StringIO(byte_str.decode('utf-8'))
 
-        except:
+        except Exception:
             log.error(traceback.format_exc())
             log.error('Could not load into string')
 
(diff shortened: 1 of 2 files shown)


mindsdb_native/libs/data_sources/file_ds.py
```diff
@@ -186,16 +188,22 @@ def _getDataIo(self, file):
             data.seek(0)
             first_few_lines = []
             i = 0
+
+            # need to have some templates to deduce a dialect
```
George3d6: What do you mean by "templates" here and bellow ? Could you give some examples.

mindsdb_native/libs/data_sources/file_ds.py
```diff
@@ -186,16 +188,22 @@ def _getDataIo(self, file):
...
+            # but it is not a good idea to deduce dialect by header row
+            # data[0]
             for line in data:
                 if line in ['\r\n','\n']:
                     continue
```
George3d6: Not added by this PR, but not exactly sure why this code has a "continue" it should have a break, i.e. this indicates a final line that only contains a newline.

mindsdb_native/libs/data_sources/file_ds.py
```diff
@@ -186,16 +188,22 @@ def _getDataIo(self, file):
...
                 first_few_lines.append(line)
                 i += 1
```
George3d6: Again, was here before, not introduced by this pr, but much cleaener to do `for i, line in enumerate(data)` than to have a manual incrementing `i`.

George3d6 (changes requested):
Please add an automated test for the failing data format to be able to confirm this is working and that we won't break the detection in the future.

mindsdb_native/libs/data_sources/file_ds.py
ilia-tsyplenkov: totally agree with this note. Done

mindsdb_native/libs/data_sources/file_ds.py
ilia-tsyplenkov: You are right, according to RFC spec CSV file couldn't contain empty lines in the middle. Fixed

mindsdb_native/libs/data_sources/file_ds.py
ilia-tsyplenkov: under template I mean `sample`
```
class csv.Sniffer
The Sniffer class is used to deduce the format of a CSV file.

The Sniffer class provides two methods:

sniff(sample, delimiters=None)¶
....
```

Issue #361, opened 2020-12-17
Flag to disable column importance determination


George3d6:
It takes a lot of time and it's also fairly useless for timeseries problems, so we should add a flag to disable it (that saves us all the time wasted on making predictions with missing columns).

PR #367, opened 2020-12-21
add disable_column_importance flag

maximlopin:
Closes https://github.com/mindsdb/mindsdb_native/issues/361

Added `disable_column_importance` flag to advanced arguments.
When it's set to `True`, column importances are not computed.

### DIFF
--- a/mindsdb_native/libs/controllers/predictor.py
+++ b/mindsdb_native/libs/controllers/predictor.py
@@ -263,7 +263,8 @@ class Predictor:
                 allow_incomplete_history = advanced_args.get('allow_incomplete_history', False),
                 quick_learn = advanced_args.get('quick_learn', None),
                 quick_predict = advanced_args.get('quick_predict', False),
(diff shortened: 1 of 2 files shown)


mindsdb_native/libs/controllers/predictor.py
```diff
@@ -263,7 +263,8 @@ def learn(self,
                 allow_incomplete_history = advanced_args.get('allow_incomplete_history', False),
                 quick_learn = advanced_args.get('quick_learn', None),
                 quick_predict = advanced_args.get('quick_predict', False),
-                apply_to_columns = advanced_args.get('apply_to_columns', {})
+                apply_to_columns = advanced_args.get('apply_to_columns', {}),
+                disable_column_importance = advanced_args.get('disable_column_importance', {})
```
ilia-tsyplenkov: since it is a boolean parameter I think that default value should be `False` not `{}`

ilia-tsyplenkov (commented):
Looks good, except one note above

PR #368, opened 2020-12-21
Never ignore group_by columns

### LINKED ISSUE #355: Force required column usage
If a column is requested as part of an argument passed to learn, such as:
* group_by
* order_by
* historical_columns

then we should make sure that we force it's usage, as in, that we don't remove it *even* if it's detected as a foreign key.

maximlopin:
Closes https://github.com/mindsdb/mindsdb_native/issues/355

- `group_by` columns are not ignored if they're also a unique identifier.

Historical columns can't be ignored because they're added to the dataframe in the `ModelInterface` phase, so no changes for that in this PR.

### DIFF
--- a/mindsdb_native/libs/phases/type_deductor/type_deductor.py
+++ b/mindsdb_native/libs/phases/type_deductor/type_deductor.py
@@ -334,7 +334,8 @@ class TypeDeductor(BaseModule):
                     if col_name not in self.transaction.lmd['predict_columns']:
                         if (self.transaction.lmd.get('tss', None) and
                                 self.transaction.lmd['tss']['is_timeseries'] and
-                                col_name in self.transaction.lmd['tss']['order_by']):
+                                col_name in self.transaction.lmd['tss']['order_by'] and
+                                col_name in self.transaction.lmd['tss']['group_by']):
                             pass
                         else:
                             self.transaction.lmd['columns_to_ignore'].append(col_name)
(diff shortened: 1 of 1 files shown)


mindsdb_native/libs/phases/type_deductor/type_deductor.py
```diff
@@ -334,7 +334,8 @@ def run(self, input_data):
                     if col_name not in self.transaction.lmd['predict_columns']:
                         if (self.transaction.lmd.get('tss', None) and
                                 self.transaction.lmd['tss']['is_timeseries'] and
-                                col_name in self.transaction.lmd['tss']['order_by']):
+                                (col_name in self.transaction.lmd['tss']['order_by'] or
+                                col_name in (self.transaction.lmd['tss']['group_by'] or []))):
```
paxcema, line 338: Why the extra `[]` at the end?

maximlopin: `self.transaction.lmd['tss']['group_by'] or []` always evaluates to an iterable, whereas `self.transaction.lmd['tss']['group_by']` can evaluate to None

paxcema: Got it, thanks

paxcema:
Looks good to me

paxcema (approved):

Issue #369, opened 2020-12-22
IndexError querying a time series with < nr_prediction data points


paxcema:
As spotted by @alejandrovillegas, train a predictor with this [dataset](https://mindsdb-web-test.s3.amazonaws.com/nyc_energy_consumption.csv) using Native:

```python
mdb = Predictor(name='bug')
mdb.learn(from_data=data,
                      to_predict='temp',
                        timeseries_settings={'order_by': ['timeStamp'],
                                           'window': 20,
                                           'use_previous_target': True,
                                           'nr_predictions': 3
                                           },
                      advanced_args={'force_predict': True,
                                     'disable_column_importance': True
                                     }
                      )
```

Then query the predictor using:

```python
mdb.predict(when_data={'timeStamp': '2012-02-07 12:00:00', 'demand': 6435.7, 'precip': 0.0})
```

Native will raise:

```
INFO:mindsdb-logger-core-logger---:/MindsDB/mindsdb_native/mindsdb_native/libs/phases/base_module.py:51 - [START] ModelInterface

ERROR:mindsdb-logger-core-logger---:/MindsDB/mindsdb_native/mindsdb_native/libs/controllers/transaction.py:181 - Could not load module ModelInterface

multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/MindsDB/mindsdb_native/mindsdb_native/libs/phases/model_interface/lightwood_backend.py", line 67, in _ts_add_previous_target
    del next_target_value_arr[del_index]
IndexError: list assignment index out of range
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/MindsDB/temp/TimeSeriesInitialDatasets/benchmark.py", line 113, in <module>
    r = mdb.predict(when_data={'timeStamp': '2012-02-07 12:00:00', 'demand': 6435.7, 'precip': 0.0})
  File "/MindsDB/mindsdb_native/mindsdb_native/libs/controllers/predictor.py", line 424, in predict
    self.transaction.run()
  File "/MindsDB/mindsdb_native/mindsdb_native/libs/controllers/transaction.py", line 325, in run
    self._call_phase_module(module_name='ModelInterface', mode='predict')
[...]

PR #370, opened 2020-12-22
Handle queries with # rows < nr_predictions for time series

paxcema:
## Why
`_ts_reshape` was failing under certain conditions, see #369 for more details.

## How
A simple fix: we iterate until the minimum between `nr_predictions` and the actual number of rows in the query.


maximlopin (approved):

maximlopin:
I also made a fix for the `window` parameter and added a small unit test: https://github.com/mindsdb/mindsdb_native/pull/371

PR #378, opened 2020-12-31
Some fixes for quick predict 

paxcema:
## Why
There were some issues when using `quick_predict` after `quick_learn`ing with a predictor.

## How
Adding explicit keyword argument passing, setting the `output_data` in the predict transaction, modifying `explain` to handle absence of confidence range info, and calling `quick_predict` in the `quick_interface` flow test.

### DIFF
--- a/mindsdb_native/libs/controllers/transaction.py
+++ b/mindsdb_native/libs/controllers/transaction.py
@@ -325,7 +325,11 @@ class PredictTransaction(Transaction):
         self._call_phase_module(module_name='ModelInterface', mode='predict')
 
         if self.lmd['return_raw_predictions']:
-            return self.hmd['predictions']
+            self.output_data = PredictTransactionOutputData(
+                transaction=self,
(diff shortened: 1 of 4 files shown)


mindsdb_native/libs/controllers/transaction.py
```diff
@@ -325,7 +325,11 @@ def run(self):
         self._call_phase_module(module_name='ModelInterface', mode='predict')
 
         if self.lmd['return_raw_predictions']:
-            return self.hmd['predictions']
+            self.output_data = PredictTransactionOutputData(
+                transaction=self,
+                data=self.hmd['predictions']
+            )
+            return self.output_data
```
George3d6, line 332: Hmh, if need be we can do it like this for now, but I think it's non-ideal, since `PredictTransactionOutputData` adds the kind of overhead we *shouldn't* have as part of the quick methods.

George3d6 (changes requested):

PR #388, opened 2021-01-14
Output class distribution flag

paxcema:
## Why
Having the belief distribution of a predictor is needed in certain use cases (see [L#354](https://github.com/mindsdb/lightwood/pull/354))

## How
Added `output_class_distribution` flag and corresponding test. Note that the feature is not yet supported for categorical tag columns, due to the way the Lightwood encoder operates.

(Note: the new test will pass once the L#354 PR is merged)

### DIFF
--- a/mindsdb_native/libs/data_types/transaction_output_row.py
+++ b/mindsdb_native/libs/data_types/transaction_output_row.py
@@ -36,7 +36,8 @@ class TransactionOutputRow:
         answers = {}
         for pred_col in self._predict_columns:
             answers[pred_col] = {}
-            prediction_row = {col: self._data[col][self._row_index] for col in self._data.keys()}
+            cols = [col for col in self._data.keys() if isinstance(self._data[col], list)]  # filter out class maps
+            prediction_row = {col: self._data[col][self._row_index] for col in cols}
 
             if self._transaction_output._transaction.lmd['tss']['is_timeseries'] and \
                     self._transaction_output._transaction.lmd['tss']['nr_predictions'] > 1:
--- a/mindsdb_native/libs/data_types/transaction_output_row.py
+++ b/mindsdb_native/libs/data_types/transaction_output_row.py
@@ -36,7 +36,7 @@ class TransactionOutputRow:
         answers = {}
         for pred_col in self._predict_columns:
             answers[pred_col] = {}
-            cols = [col for col in self._data.keys() if 'class_map' not in col]  # filter out class maps
+            cols = [col for col in self._data.keys() if '_class_distribution' not in col]
             prediction_row = {col: self._data[col][self._row_index] for col in cols}
 
             if self._transaction_output._transaction.lmd['tss']['is_timeseries'] and \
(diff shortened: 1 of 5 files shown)


mindsdb_native/libs/data_types/transaction_output_row.py
```diff
@@ -36,7 +36,8 @@ def explain(self):
         answers = {}
         for pred_col in self._predict_columns:
             answers[pred_col] = {}
-            prediction_row = {col: self._data[col][self._row_index] for col in self._data.keys()}
+            cols = [col for col in self._data.keys() if 'class_map' not in col]  # filter out class maps
```
George3d6: I'm confused, shouldn't this be `'_class_distribution' not in col` ?

mindsdb_native/libs/phases/model_interface/lightwood_backend.py
```diff
@@ -552,10 +558,12 @@ def predict(self, mode='predict', ignore_columns=None, all_mixers=False):
 
             predictions = self.predictor.predict(when_data=run_df)
 
-
             if self.transaction.lmd['quick_predict']:
                 for k in predictions:
                     formated_predictions[k] = predictions[k]['predictions']
+                    if 'class_distribution' in predictions[k]:
```
George3d6: Better to check `if self.transaction.lmd['output_class_distribution']` maybe, to avoid certain silent failures.

mindsdb_native/libs/phases/model_interface/lightwood_backend.py
```diff
@@ -564,6 +572,9 @@ def predict(self, mode='predict', ignore_columns=None, all_mixers=False):
                     continue
 
                 formated_predictions[k] = predictions[k]['predictions']
+                if 'class_distribution' in predictions[k]:
```
George3d6: Better to check `if self.transaction.lmd['output_class_distribution']` maybe, to avoid certain silent failures.

tests/unit_tests/libs/controllers/test_predictor.py
```diff
@@ -12,6 +12,7 @@
 from sklearn import preprocessing
 from sklearn.linear_model import LinearRegression
 from sklearn.metrics import r2_score, f1_score, accuracy_score
+from sklearn.datasets import load_iris
```
George3d6, line 15: Hmh, never thought about using datasets from sklearn, they do pose some issues with e.g. data being pre encoded I believe, but in this case I guess it works just fine.

tests/unit_tests/libs/controllers/test_predictor.py
```diff
@@ -484,3 +485,21 @@ def test_empty_column(self):
...
+            from_data=df,
+            to_predict='target',
+            stop_training_in_x_seconds=1,
+            use_gpu=False,
+            advanced_args={'debug': True, 'output_class_distribution': True}
+        )
+
+        results = mdb.predict(when_data=data.data.iloc[[0]])
+        assert 'target_class_distribution' in results._data
+        assert np.isclose(np.sum(results._data['target_class_distribution']), 1)
```
George3d6: Add an assert of the nr of labels, i.e. the length of every member of the array under `target_class_distribution`, since this is presumably known for the training data and should be the same for every member.

George3d6 (changes requested):

mindsdb_native/libs/data_types/transaction_output_row.py
paxcema: Yup, nice catch, this is leftover from a previous commit.

PR #408, opened 2021-01-27
removed failure

George3d6:
Models no longer allowed to fail for a given column


George3d6:
Part of the lightgbm PR

PR #431, opened 2021-02-03
Fixes for mongo atlas connection

StpMax:
**why**

Alejo had `SSL: CERTIFICATE_VERIFY_FAILED` error when tried connect to mongo atlas.

**how**

We tried many ways to fix issue, looks like this case work. Here is theme on mongo forum:
https://developer.mongodb.com/community/forums/t/cant-connect-to-atlas-on-macos/15009/2

### DIFF
--- a/mindsdb_native/libs/data_sources/mongodb_ds.py
+++ b/mindsdb_native/libs/data_sources/mongodb_ds.py
@@ -30,20 +30,29 @@ class MongoDS(DataSource):
     def query(self, q):
         assert isinstance(q, dict)
 
-        if self.user == '':
-            conn = MongoClient(
-                host=self.host,
-                port=self.port,
-                tlsCAFile=certifi.where()
-            )
-        else:
-            conn = MongoClient(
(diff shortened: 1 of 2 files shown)


mindsdb_native/libs/data_sources/mongodb_ds.py
```diff
@@ -29,18 +30,30 @@ def __init__(self, query, collection, database='database',
...
-            )
+        kwargs = {}
+
+        if isinstance(self.user, str) and len(self.user) > 0:
+            kwargs['username'] = self.user
+
+        if isinstance(self.password, str) and len(self.password) > 0:
+            kwargs['password'] = self.password
+
+        if re.match(r'\/\?.*tls=true', self.host.lower()):
```
ilia-tsyplenkov, line 41: host.lower().endswith('tls=true') looks a little bit simpler, or not?

ilia-tsyplenkov (changes requested):

PR #436, opened 2021-02-04
Boosted absolute error for confidence in regression tasks

paxcema:
## Why
As discovered by @ZoranPandovski, the width of prediction regions could vary wildly for the same dataset when training several predictors, which would lead to very different reported accuracies even when the predictors themselves were nearly identical in predictive capabilities. 

The issue stems from using the signed error function for nonconformity scores, which could generate non-monotonically decreasing interval widths (i.e. there was no guarantee that less confidence would mean a tighter spread). Switching to the absolute error function solves this issue, and grants stability to the width in the predicted region.

## How
The PR changes the nonconformity function used in regression to a "boosted" variant of Nonconformist's `AbsErrorErrFunc`, which interpolates the nonconf scores whenever our validation dataset has less than 100 samples. For increased precision when setting the reported accuracy, it additionally reverts to checking all possible significance levels  because optimizations made in #377 imply a much cheaper evaluation here.

### DIFF
--- a/mindsdb_native/libs/helpers/conformal_helpers.py
+++ b/mindsdb_native/libs/helpers/conformal_helpers.py
@@ -39,8 +39,7 @@ def get_conf_range(X, icp, target, typing_info, lmd, std_tol=1):
         all_ranges = icp.predict(X.values)
 
         # iterate over confidence levels until spread >= a multiplier of the dataset stddev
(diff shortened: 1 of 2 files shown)


George3d6:
@paxcema why were signed errors used to being with? Just a mistake?

PR #446, opened 2021-02-11
Add 'limit' and 'sort' option for mongo DS

ilia-tsyplenkov:
Implements [#1097](https://github.com/mindsdb/mindsdb/issues/1097) on mindsdb_native side
How:
 - by calling `sort` and/or `limit` methods of object returned by `find`

### DIFF
--- a/mindsdb_native/libs/data_sources/mongodb_ds.py
+++ b/mindsdb_native/libs/data_sources/mongodb_ds.py
@@ -8,20 +8,28 @@ from pandas.api.types import is_numeric_dtype
 from mindsdb_native.libs.data_types.data_source import DataSource
 
 class MongoDS(DataSource):
-    def __init__(self, query, collection, database='database',
+    def __init__(self, query, collection,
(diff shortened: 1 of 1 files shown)


mindsdb_native/libs/data_sources/mongodb_ds.py
```diff
@@ -8,20 +8,28 @@
 from mindsdb_native.libs.data_types.data_source import DataSource
 
 class MongoDS(DataSource):
-    def __init__(self, query, collection, database='database',
+    def __init__(self, query, collection,
+                 database='database', limit=None, sort_by=None,
                  host='127.0.0.1', port=None, user=None, password=None):
```
George3d6: don't change argument order, add the new ones at the end.

George3d6 (changes requested):

mindsdb_native/libs/data_sources/mongodb_ds.py
ilia-tsyplenkov: Fixed

PR #456, opened 2021-02-23
Max processes advanced arg

paxcema:
## Why
To give advanced users more control over how many processes to spawn and/or how much base memory each worker would use. 

## How
Adding the respective arguments and modifying the `mp_helper.get_nr_procs` method a bit.


George3d6 (approved):
Looks good :)

PR #468, opened 2021-03-17
Anomaly detection for numerical time series

paxcema:
## Why
Add anomaly detection support for numerical time series (see [M#1116](https://github.com/mindsdb/mindsdb/issues/1116)).

This first working iteration uses an approach based on confidence ranges that the conformal predictor outputs. Any incoming true value that falls outside the predicted bounds will be flagged as an anomaly under the `output_data[f'{target}_anomaly]` boolean field in the `PredictTransaction`.

Anomaly detection is enabled by default (relevant argument is `anomaly_detection=True` in `advanced_args` for `Predictor.predict()`).

To customize the sensitivity to each particular use case, two additional parameters are exposed in `advanced_args`:
1. `anomaly_error_rate` (float): forces an error rate in the ICP confidence estimator. This directly impacts the sensitivity of anomaly detection. Default behavior (`None`) determines this automatically.
2. `anomaly_cooldown` (int): number of timesteps that will be flagged as an anomaly after a triggering event. This is to compensate the predictor response, and assumes a series is regularly spaced. 

## How
Given the confidence-based approach, the implementation is fairly straightforward by flagging in `PredictTransaction` any real value that is not within bounds. Some minor refactoring was needed to pass the additional arguments as well.

### DIFF
--- a/mindsdb_native/libs/controllers/predictor.py
+++ b/mindsdb_native/libs/controllers/predictor.py
@@ -61,9 +61,6 @@ def _prepare_timeseries_settings(user_provided_settings):
         use_previous_target=True,
         nr_predictions=1,
         historical_columns=[],
-        anomaly_detection=True,
-        anomaly_error_rate=None,  # (None or float) forces specific confidence level in ICP
-        anomaly_cooldown=1,  # (Int) in time steps; implicitly assumes series are regularly spaced
     )
 
     if len(user_provided_settings) > 0:
@@ -431,7 +428,12 @@ class Predictor:
                 use_database_history = advanced_args.get('use_database_history', False),
                 allow_incomplete_history = advanced_args.get('allow_incomplete_history', False),
                 quick_predict = advanced_args.get('quick_predict', False),
-                return_raw_predictions = advanced_args.get('return_raw_predictions', False)
+                return_raw_predictions = advanced_args.get('return_raw_predictions', False),
+                anomaly_detection = advanced_args.get('anomaly_detection', True),
+                # (None or float) forces specific confidence level in ICP
+                anomaly_error_rate = advanced_args.get('anomaly_error_rate', None),
+                # (Int) in time steps; implicitly assumes series are regularly spaced
+                anomaly_cooldown = advanced_args.get('anomaly_cooldown', 1),
             )
 
             self.transaction = PredictTransaction(
--- a/mindsdb_native/libs/controllers/predictor.py
+++ b/mindsdb_native/libs/controllers/predictor.py
@@ -54,14 +54,16 @@ def _prepare_sample_settings(user_provided_settings,
 
 def _prepare_timeseries_settings(user_provided_settings):
     timeseries_settings = dict(
-        is_timeseries=False
-        ,group_by=None
-        ,order_by=None
-        ,window=None
-        ,use_previous_target=True
-        ,nr_predictions=1
-        ,historical_columns=[]
-        ,anomaly_detection=True
+        is_timeseries=False,
+        group_by=None,
+        order_by=None,
+        window=None,
+        use_previous_target=True,
+        nr_predictions=1,
+        historical_columns=[],
+        anomaly_detection=True,
+        anomaly_error_rate=None,  # (None or float) forces specific confidence level in ICP
+        anomaly_cooldown=1,  # (Int) in time steps; implicitly assumes series are regularly spaced
     )
 
     if len(user_provided_settings) > 0:
@@ -71,7 +73,6 @@ def _prepare_timeseries_settings(user_provided_settings):
             raise Exception(f'Invalid timeseries settings, you must specify a window size')
         else:
             timeseries_settings['is_timeseries'] = True
-
 
     for k in user_provided_settings:
         if k in timeseries_settings:
--- a/mindsdb_native/libs/controllers/predictor.py
+++ b/mindsdb_native/libs/controllers/predictor.py
@@ -61,6 +61,7 @@ def _prepare_timeseries_settings(user_provided_settings):
         ,use_previous_target=True
         ,nr_predictions=1
         ,historical_columns=[]
+        ,anomaly_detection=True
     )
 
     if len(user_provided_settings) > 0:
--- a/mindsdb_native/libs/controllers/predictor.py
+++ b/mindsdb_native/libs/controllers/predictor.py
@@ -58,9 +58,9 @@ def _prepare_timeseries_settings(user_provided_settings):
         group_by=None,
         order_by=None,
         window=None,
-        use_previous_target=True,
-        nr_predictions=1,
-        historical_columns=[],
+        use_previous_target=True,  # adds target previous values to the TS encoder input data
+        nr_predictions=1,          # forecasting window, instances nr_predictions columns by displacing each target
+        historical_columns=[],     # each of these gets encoded by its own TS encoder
     )
 
     if len(user_provided_settings) > 0:
@@ -430,9 +430,13 @@ class Predictor:
                 quick_predict = advanced_args.get('quick_predict', False),
                 return_raw_predictions = advanced_args.get('return_raw_predictions', False),
                 anomaly_detection = advanced_args.get('anomaly_detection', True),
+
                 # (None or float) forces specific confidence level in ICP
                 anomaly_error_rate = advanced_args.get('anomaly_error_rate', None),
-                # (Int) in time steps; implicitly assumes series are regularly spaced
+
+                # (Int) ignores anomaly detection for N steps after an
+                # initial anomaly triggers the cooldown period;
+                # implicitly assumes series are regularly spaced
                 anomaly_cooldown = advanced_args.get('anomaly_cooldown', 1),
             )
 
--- a/mindsdb_native/libs/controllers/transaction.py
+++ b/mindsdb_native/libs/controllers/transaction.py
@@ -478,6 +478,14 @@ class PredictTransaction(Transaction):
                     confs = [[a, b] for a, b in zip(result['lower'], result['upper'])]
                     output_data[f'{predicted_col}_confidence_range'] = confs
 
+                    # anomaly detection
+                    if self.lmd['tss']['is_timeseries'] and self.lmd['tss']['anomaly_detection']:
+                        anomalies = []
+                        for (l, u), t in zip(output_data[f'{predicted_col}_confidence_range'],
+                                             output_data[f'__observed_{predicted_col}']):
+                            anomalies.append(not l <= t <= u)
+                        output_data[f'{predicted_col}_anomaly'] = anomalies
+
         else:
             for predicted_col in self.lmd['predict_columns']:
                 output_data[f'{predicted_col}_confidence'] = [None] * len(output_data[predicted_col])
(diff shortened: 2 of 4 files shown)


mindsdb_native/libs/controllers/predictor.py
```diff
@@ -54,13 +54,13 @@ def _prepare_sample_settings(user_provided_settings,
...
 def _prepare_timeseries_settings(user_provided_settings):
     timeseries_settings = dict(
-        is_timeseries=False
-        ,group_by=None
-        ,order_by=None
-        ,window=None
-        ,use_previous_target=True
-        ,nr_predictions=1
-        ,historical_columns=[]
+        is_timeseries=False,
```
hakunanatasha, line 57: ,what , you dont like ,front commas? (JKJK)

mindsdb_native/libs/controllers/predictor.py
```diff
@@ -54,13 +54,13 @@ def _prepare_sample_settings(user_provided_settings,
 
 def _prepare_timeseries_settings(user_provided_settings):
```
hakunanatasha, line 55: up to preference but can you put a docstring here to explain what each of these vars are doing, nr_predictions is not obvious to me but the others are self-evident

mindsdb_native/libs/controllers/predictor.py
```diff
@@ -429,7 +428,12 @@ def predict(self,
                 use_database_history = advanced_args.get('use_database_history', False),
                 allow_incomplete_history = advanced_args.get('allow_incomplete_history', False),
                 quick_predict = advanced_args.get('quick_predict', False),
-                return_raw_predictions = advanced_args.get('return_raw_predictions', False)
+                return_raw_predictions = advanced_args.get('return_raw_predictions', False),
+                anomaly_detection = advanced_args.get('anomaly_detection', True),
```
hakunanatasha, line 432: is anomally_detection triggered by default?

mindsdb_native/libs/controllers/transaction.py
```diff
@@ -381,6 +381,10 @@ def run(self):
                                   DATA_TYPES.CATEGORICAL in typing_info['data_type_dist'].keys())) and \
                                   typing_info['data_subtype'] != DATA_SUBTYPES.TAGS
 
+                is_anomaly_task = is_numerical and \
```
hakunanatasha, line 384: is it better to assert(is_numerical)?

mindsdb_native/libs/controllers/transaction.py
```diff
@@ -427,8 +431,11 @@ def run(self):
 
                     # convert (B, 2, 99) into (B, 2) given width or error rate constraints
                     if is_numerical:
-                        significances, confs = get_numerical_conf_range(all_confs, predicted_col,
-                                                                        self.lmd['stats_v2'])
+                        error_rate = self.lmd['anomaly_error_rate'] if is_anomaly_task else None
```
hakunanatasha, line 434: what is the behavior when you pass Nones?

mindsdb_native/libs/controllers/transaction.py
```diff
@@ -458,9 +465,11 @@ def run(self):
                                 # predict and get confidence level given width or error rate constraints
                                 if is_numerical:
```
hakunanatasha, line 466: why is there 2 is_numericals? I'm missing the flow

hakunanatasha: i should add its probably fine for now, but likely can be streamlined as I see it in a few places

mindsdb_native/libs/controllers/transaction.py
```diff
@@ -458,9 +465,11 @@ def run(self):
...
                                     all_confs = icps[frozenset(group)].predict(X.values)
+                                    error_rate = self.lmd['anomaly_error_rate'] if is_anomaly_task else None
                                     significances, confs = get_numerical_conf_range(all_confs, predicted_col,
                                                                                     self.lmd['stats_v2'],
-                                                                                    group=frozenset(group))
+                                                                                    group=frozenset(group),
+                                                                                    error_rate=error_rate)
                                     result.loc[X.index, 'lower'] = confs[:, 0]
```
hakunanatasha, line 473: is results always in the form of a dataframe?

mindsdb_native/libs/helpers/confidence_helpers.py
```diff
@@ -66,39 +66,49 @@ def set_conf_range(X, icp, target, typing_info, lmd, std_tol=1, group='__default
...
-            significance = (99 - idx) / 100
-            diff = sample[1, idx] - sample[0, idx]
-            if diff <= tolerance:
-                conf_range = list(sample[:, idx])
-                if stats[predicted_col].get('positive_domain', False):
-                    conf_range[0] = max(0.0, conf_range[0])
-
-                significances.append(significance)
+def get_numerical_conf_range(all_confs, predicted_col, stats, std_tol=1, group='__default', error_rate=None):
+    """ Gets prediction bounds for numerical targets, based on ICP estimation and width tolerance
```
hakunanatasha, line 70: is it worth type hinting, if at least in the doc string, here?

mindsdb_native/libs/helpers/confidence_helpers.py
```diff
@@ -66,39 +66,49 @@ def set_conf_range(X, icp, target, typing_info, lmd, std_tol=1, group='__default
...
+        conf = 1 - error_rate
+        conf_idx = int(100*error_rate) - 1
+        conf_ranges = all_confs[:, :, conf_idx]
+        significances = [conf for _ in range(conf_ranges.shape[0])]
 
-            conf_ranges.append(conf_range)
-
-    return significances, np.array(conf_ranges)
+    if stats[predicted_col].get('positive_domain', False):
+        conf_ranges[conf_ranges < 0] = 0
```
hakunanatasha, line 118: I see youve extended the confidence ranges - since it looks like you're looking for all instances where conf_ranges < 0 to convert to 0

mindsdb_native/libs/helpers/confidence_helpers.py
```diff
@@ -66,39 +66,49 @@ def set_conf_range(X, icp, target, typing_info, lmd, std_tol=1, group='__default
...
                 conf_ranges.append(conf_range)
-                break
-        else:
-            significances.append(0.9991)  # default: confident that value falls inside big bounds
-            bounds = sample[:, 0]
-            sigma = (bounds[1] - bounds[0]) / 4
-            conf_range = [bounds[0] - sigma, bounds[1] + sigma]
 
-            if stats[predicted_col].get('positive_domain', False):
-                conf_range[0] = max(0.0, conf_range[0])
```
hakunanatasha, line 97: was there a length change? I'm confused why the first item here is the only one that gets modified

mindsdb_native/libs/helpers/confidence_helpers.py
```diff
@@ -66,39 +66,49 @@ def set_conf_range(X, icp, target, typing_info, lmd, std_tol=1, group='__default
...
+        conf_ranges = np.array(conf_ranges)
+    else:
+        # fixed error rate
+        conf = 1 - error_rate
+        conf_idx = int(100*error_rate) - 1
```
hakunanatasha, line 113: error_rate is a float but also needs to be less than 1; this should be explicitly checked in an assertion or statement? since most are handled internally maybe not, but what happens if I make error rate negative or something weird.

mindsdb_native/libs/helpers/confidence_helpers.py
hakunanatasha: or rather, how large can conf_ranges be?

mindsdb_native/libs/helpers/confidence_helpers.py
```diff
@@ -115,3 +125,27 @@ def get_categorical_conf(all_confs, conf_candidates):
         else:
             significances.append(0.005)  # default: not confident label is the predicted one
     return significances
+
+
+def get_anomalies(bounds, observed_series, cooldown=1):
```
hakunanatasha, line 141: what is the role of cooldown? It looks like when an anomally triggers you reset the counter

mindsdb_native/libs/controllers/predictor.py
paxcema: Hah! This is the linter doing its thing

mindsdb_native/libs/controllers/predictor.py
paxcema: Yes, no reason other than it is very cheap to compute

mindsdb_native/libs/controllers/transaction.py
paxcema: It automatically finds a "tight enough" bound based on the target standard deviation seen at training time

mindsdb_native/libs/helpers/confidence_helpers.py
paxcema: Depending on the nonconformity scores they can be quite big. Also, unless normalized they are symmetrical, which is why for a large enough bound you can get a partially negative prediction range for a target you know cannot be negative. That's why this filter is in place.

mindsdb_native/libs/helpers/confidence_helpers.py
paxcema: It's based on an implicit guarantee, I will document this on the code itself: if `positive_domain` flag is enabled, you can be certain predictions are >= 0. This in turn means only the lower bound can ever be negative.

Still, this has been rewritten as `conf_ranges[conf_ranges < 0] = 0`, hopefully a little bit clearer.

mindsdb_native/libs/helpers/confidence_helpers.py
paxcema: Good question, I'll add this in the code too.

Cooldown is a grace period during which the predictor will be coping with the dynamics introduced by the anomaly, so we avoid additional flagging in case the series has gone back to its normal dynamics but the predictor hasn't yet. This is useful for spike-type anomalies, maybe not so for gradual ones.

mindsdb_native/libs/controllers/transaction.py
paxcema: Hmm probably. So, the flow is that we get the base confidence for all predictions, and then check whether we have groups that will refine those bounds (like in grouped time series). So in essence it's part of the same process, just slightly different. 

I will think how to streamline it, but I agree it's out of the scope of this PR

mindsdb_native/libs/helpers/confidence_helpers.py
hakunanatasha: does this require the user to have a notion of whether they have spike or gradual anomalies? (i.e. maybe like a detector - a spike can be a 'recognized signal' versus a gradual one a temperature signal being a slow loss of temperature over time)

mindsdb_native/libs/controllers/transaction.py
paxcema: Yes, and it's done to cope with the tricky issue of knowing what subset of rows in the query data belong to which group combination through the DF index. 

It's cast back into lists at the end of the procedure, so the DF is just for internal use.

mindsdb_native/libs/controllers/predictor.py
paxcema: Will do

mindsdb_native/libs/controllers/transaction.py
paxcema: doubtful, we don't want an exception if `not is_numerical`

mindsdb_native/libs/helpers/confidence_helpers.py
paxcema: Yup! And this being a first iteration, we will have to explore how to best tackle gradual anomalies.

mindsdb_native/libs/helpers/confidence_helpers.py
paxcema: Very good point, I'll add the assertion

mindsdb_native/libs/helpers/confidence_helpers.py
paxcema: Agreed

PR #476, opened 2021-04-01
Store ICP groups with str keys

paxcema:
## Why
Solves [M#1186](https://github.com/mindsdb/mindsdb/issues/1186).

Grouped time series predictors were not being saved due to an error in SQL Alchemy, as it could not use `frozensets` as keys. These `frozensets` were used to store ICPs for each different group. This PR changes those keys to a simple `'_'.join()` of the values in the grouped columns for each possible group.

Note: ignore deceiving branch name


George3d6 (changes requested):
Not the way to go, see slack thread for details.

Issue #471, opened 2021-03-26
Failed to `order_by` date error


ZoranPandovski:
Link to download dataset | https://www.kaggle.com/datasnaek/youtube-new
-- | --
Steps to Reproduce | Upload the Dataset, target variable is Views, order by `trending date`
 

Error stack trace:

```
C:\PROGRA~1\MindsDB\python\lib\site-packages\mindsdb_native\libs\phases\model_interface\lightwood_backend.py", line 
125, in _ts_reshape row[col] = float(row[col]) ValueError: could not convert string to float: '17.01.12' During handling of the
 above exception, another exception occurred: Traceback (most recent call last): File "C:\PROGRA~1\MindsDB\python\lib\site-packages\mindsdb_native\libs\controllers\transaction.py", line 246, in _run self._call_phase_module(module_name='ModelInterface', mode='train') File "C:\PROGRA~1\MindsDB\python\lib\site-
packages\mindsdb_native\libs\controllers\transaction.py", line 178, in _call_phase_module ret = module(self.session, self)
(**kwargs) File "C:\PROGRA~1\MindsDB\python\lib\site-packages\mindsdb_native\libs\phases\base_module.py", line 53, in
 __call__ ret = self.run(**kwargs) File "C:\PROGRA~1\MindsDB\python\lib\site-
packages\mindsdb_native\libs\phases\model_interface\model_interface.py", line 18, in run self.transaction.model_backend.train()
 File "C:\PROGRA~1\MindsDB\python\lib\site-packages\mindsdb_native\libs\phases\model_interface\lightwood_backend.py", line 337, in train train_df, secondary_type_dict, _, train_df_gb_map = self._ts_reshape(self.transaction.input_data.train_df) File 
"C:\PROGRA~1\MindsDB\python\lib\site-packages\mindsdb_native\libs\phases\model_interface\lightwood_backend.py", line 
127, in _ts_reshape raise ValueError(f'Failed to order based on column: "{col}" due to faulty value: {row[col]}') ValueError: Failed 
to order based on column: "trending_date" due to faulty value: 17.01.12
```

PR #480, opened 2021-04-05
Support multiple date/date+time formats

maximlopin:
Closes https://github.com/mindsdb/mindsdb_native/issues/471
Closes https://github.com/mindsdb/mindsdb_native/issues/481

- Detect multiple date/date+time formats
- Allow user to provide list of formats in advanced_args
- Add unit test

### DIFF
--- a/mindsdb_native/libs/phases/data_transformer/data_transformer.py
+++ b/mindsdb_native/libs/phases/data_transformer/data_transformer.py
@@ -24,13 +24,13 @@ def _try_round(x):
         return None
 
 
-def _standardize_date(date_str):
+def _standardize_date(date_str, stats_v2):
     try:
         # will return a datetime object
-        date = parse_datetime(date_str)
+        date = datetime.datetime.strptime(date_str, additional_info['datetime_formats']['date_fmt'])
     except Exception:
         try:
-            date = datetime.datetime.utcfromtimestamp(date_str)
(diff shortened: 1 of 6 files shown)


mindsdb_native/libs/phases/data_transformer/data_transformer.py
```diff
@@ -24,29 +24,30 @@ def _try_round(x):
         return None
 
 
-def _standardize_date(date_str):
+def _standardize_date(date_str, fmt):
     try:
         # will return a datetime object
-        date = parse_datetime(date_str)
```
paxcema, line 30: I believe `parse_datetime` is no longer used and so we can remove its import

mindsdb_native/libs/phases/type_deductor/type_deductor.py
```diff
@@ -185,7 +195,7 @@ def get_column_data_type(arg_tup, lmd):
     if known_type_dist:
         max_known_dtype, max_known_dtype_count = max(
             known_type_dist.items(),
-            key=lambda kv: kv[0]
+            key=lambda kv: kv[1]
```
paxcema, line 198: Why the index change here? I'm not sure I follow

maximlopin: because the max() should be on values, IDK why it used to be keys :/

PR #484, opened 2021-04-12
Infer time series rows for streaming

paxcema:
## Why

For stream usecases, we need to predict for a datapoint that is yet to be published by the stream.

## How

As detailed in the related issue [M#1171](https://github.com/mindsdb/mindsdb/issues/1171), specifying `make_predictions==False` for the predict DataFrame will trigger the inferring policy. At the moment, it extracts the temporal data for the `order_by` column by comparing the last two entries in the stream cache. This could be changed to consider the complete cache, or even previous values. On the other hand, the values for the rest of the columns is carried over from the last row (here we could explore imputing techniques, basically treating the row as missing data and generating values conditioned on the cache).


tests/unit_tests/libs/controllers/test_predictor_timeseries.py
```diff
@@ -318,3 +318,67 @@ def test_split_models(self):
...
+
+        features = generate_value_cols(['date', 'int', 'int', 'true'], data_len, ts_hours * 3600)
+        features[-1][0] = 'make_predictions'  # add make_predictions column as mindsdb would
+        labels = [generate_timeseries_labels(features[:-1])]
+
+        feature_headers = list(map(lambda col: col[0], features))
+        label_headers = list(map(lambda col: col[0], labels))
+
+        # Create the training dataset and save it to a file
+        columns_train = list(map(lambda col: col[1:int(len(col) * 3 / 4)], features))
```
ilia-tsyplenkov, line 336: Is `make_predictions` column required for train data? If `yes` I don't think that it will be good for users. As I understand that `make_predictions` column needs for TS prediction only, not for training.

ilia-tsyplenkov (changes requested):

tests/unit_tests/libs/controllers/test_predictor_timeseries.py
paxcema: That's a good question. It is not needed in train data. Notice in this test the last column of the dataset (which is "useless" as it contains True values only) is replaced in the `test_file_name` case by a `make_predictions = False` column, which complies with the behavior you expect: only in `predict()` call do we need `make_predictions`.

mindsdb_native/libs/controllers/transaction.py
```diff
@@ -328,11 +328,16 @@ def run(self):
...
         if 'make_predictions' in self.input_data.data_frame.columns:
-            predictions_df = pd.DataFrame(
-                self.input_data.data_frame[
-                    self.input_data.data_frame['make_predictions'] == True
-                ]
-            )
+            to_predict = self.input_data.data_frame[self.input_data.data_frame['make_predictions'] == True]
+            if to_predict.shape[0] == 0:
+                # assume infer mode
+                predictions_df = self.input_data.data_frame
```
George3d6: shouldn't this be just the last row for each group by value?

mindsdb_native/libs/phases/model_interface/lightwood_backend.py
```diff
@@ -86,16 +99,21 @@ def __init__(self, transaction):
...
+        if 'make_predictions' in original_df.columns:
+            index = original_df[original_df['make_predictions'].map({'True': True, 'False': False, True: True, False: False}) == True]
+            infer_mode = index.shape[0] == 0  # condition to trigger: make_predictions is set to False everywhere
+        else:
+            infer_mode = False
         original_index_list = []
         idx = 0
         for row in original_df.itertuples():
-            if _make_pred(row):
+            if _make_pred(row) or infer_mode:
```
George3d6, line 117: This seems wrong (?), i.e. in infer mode we want to return a sinlge row, the infered row. This is adding every single "history" row we get to the list of indexes of rows that will be returned.

mindsdb_native/libs/phases/model_interface/lightwood_backend.py
```diff
@@ -141,6 +159,13 @@ def _ts_reshape(self, original_df):
         else:
             df_arr = [original_df]
 
+        last_index = original_df['original_index'].max()
+        for i, subdf in enumerate(df_arr):
+            if 'make_predictions' in subdf.columns and mode == 'predict':
+                if infer_mode:
+                    df_arr[i] = _ts_infer_next_row(subdf, ob_arr, last_index)
```
George3d6, line 167: I think this will not have the intended behavior, in part see comment above, in part see the fact that this "new" infered dataframe will never be appended to the "original" data which native will "merge" with the timeseries predictions (based on the `original_index_list`).

So we won't be returning the "correct" values the prediction is made for to the user.

tests/unit_tests/libs/controllers/test_predictor_timeseries.py
```diff
@@ -318,3 +318,67 @@ def test_split_models(self):
...
+        columns_train.extend(list(map(lambda col: col[1:int(len(col) * 3 / 4)], labels)))
+        columns_to_file(
+            columns_train,
+            train_file_name,
+            headers=[*feature_headers, *label_headers]
+        )
+
+        # force make_predictions column to be false, thus triggering inference for stream use cases
+        features[-1] = generate_value_cols(['false'], data_len, ts_hours * 3600)[0]
+        features[-1][0] = 'make_predictions'
```
George3d6, line 346: This doesn't seem to achieve the intended behaviour, i.e. it doesn't force "make_predictions" to `False`.

tests/unit_tests/utils.py
```diff
@@ -84,6 +96,12 @@ def generate_value_cols(types, length, ts_period=48*3600):
             columns[-1].extend(
                 generate_timeseries(length=length, _type=t, period=ts_period))
             continue
+        elif t == 'bool':
+            gen_fun = gen_bool
```
George3d6: `gen_fun = lambda: bool(random.randint(0, 1))`

tests/unit_tests/utils.py
```diff
@@ -84,6 +96,12 @@ def generate_value_cols(types, length, ts_period=48*3600):
...
+        elif t == 'true':
+            gen_fun = gen_bool_true
```
George3d6: `gen_fun = lambda: True`

tests/unit_tests/utils.py
```diff
@@ -84,6 +96,12 @@ def generate_value_cols(types, length, ts_period=48*3600):
...
+        elif t == 'false':
+            gen_fun = gen_bool_false
```
George3d6: `gen_fun = lambda: False`

tests/unit_tests/utils.py
```diff
@@ -69,6 +69,18 @@ def rand_float():
...
+def gen_bool():
+    return bool(random.randint(0, 1))
+
+
+def gen_bool_true():
+    return True
+
+
+def gen_bool_false():
+    return False
```
George3d6: delete, see bellow

George3d6 (changes requested):

tests/unit_tests/libs/controllers/test_predictor_timeseries.py
paxcema: Hmm, I've just checked and it does the trick for the test data frame all the way through to the ModelInterface

mindsdb_native/libs/controllers/transaction.py
paxcema: The cached pred DF has the last row for each group now

mindsdb_native/libs/phases/model_interface/lightwood_backend.py
paxcema: Noted. I've now opted for setting `original_index` to None in inferred rows and slightly modifying the logic to generate `timeseries_row_mapping`

ilia-tsyplenkov (approved):
It works for covid dataset.

George3d6 (approved):

Issue #492, opened 2021-04-20
eeg_eyes dataset: could not convert string to float


maximlopin:
```
Traceback (most recent call last):
  File "C:\Users\maxim\Envs\mindsdb\lib\site-packages\mindsdb_native\libs\phases\model_interface\lightwood_backend.py", line 133, in _ts_reshape
    row[col] = float(row[col])
ValueError: could not convert string to float: '2001-09-03 08:00:00.000000'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "Q:\Desktop\mdb\private-benchmarks\benchmarks\run.py", line 230, in run
    use_cache=False if args.force_run else True
  File "Q:\Desktop\mdb\private-benchmarks\benchmarks\frameworks\frameworks.py", line 40, in run
    result = self._run(ds, df_train, df_test)
  File "Q:\Desktop\mdb\private-benchmarks\benchmarks\frameworks\frameworks.py", line 208, in _run
    learn_kwargs=ds['learn_kwargs']
  File "Q:\Desktop\mdb\private-benchmarks\benchmarks\mdb\benchmark.py", line 22, in run
    predictor.learn(from_data=df_train, to_predict=target, **learn_kwargs)
  File "C:\Users\maxim\Envs\mindsdb\lib\site-packages\mindsdb_native\libs\controllers\predictor.py", line 315, in learn
    self.transaction.run()
  File "C:\Users\maxim\Envs\mindsdb\lib\site-packages\mindsdb_native\libs\controllers\transaction.py", line 269, in run
    self._run()
  File "C:\Users\maxim\Envs\mindsdb\lib\site-packages\mindsdb_native\libs\controllers\transaction.py", line 265, in _run
    raise e
  File "C:\Users\maxim\Envs\mindsdb\lib\site-packages\mindsdb_native\libs\controllers\transaction.py", line 238, in _run
    self._call_phase_module(module_name='ModelInterface', mode='train')
  File "C:\Users\maxim\Envs\mindsdb\lib\site-packages\mindsdb_native\libs\controllers\transaction.py", line 170, in _call_phase_module
    ret = module(self.session, self)(**kwargs)
  File "C:\Users\maxim\Envs\mindsdb\lib\site-packages\mindsdb_native\libs\phases\base_module.py", line 53, in __call__
    ret = self.run(**kwargs)
[...]

PR #495, opened 2021-04-20
better date subtype detection

maximlopin:
Closes https://github.com/mindsdb/mindsdb_native/issues/493
Closes https://github.com/mindsdb/mindsdb_native/issues/492

- Add  `'%Y-%m-%d %H:%M:%S.%f'` and `'%m/%d/%y'` to list of supported date formats
- Speed up date subtype detection (fist check if at least one format matches at least one cell, then iterate through all cells and all formats and see which format matches the most cells, and based on which format has the most matches return the subtype [`DATA_SUBTYPES.TIMESTAMP` or `DATA_SUBTYPES.DATE`])

### DIFF
--- a/mindsdb_native/libs/controllers/predictor.py
+++ b/mindsdb_native/libs/controllers/predictor.py
@@ -271,8 +271,6 @@ class Predictor:
                 remove_columns_with_missing_targets = advanced_args.get('remove_columns_with_missing_targets', True),
                 max_processes = advanced_args.get('max_processes', None),
                 max_per_proc_usage = advanced_args.get('max_per_proc_usage', None),
-                date_fmts = advanced_args.get('date_fmts', None),
-                datetime_fmts = advanced_args.get('datetime_fmts', None),
 
                 learn_started_at = time.time(),
             )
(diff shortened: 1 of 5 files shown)


George3d6 (commented):
I'm a bit confused by this to be honest. Namely, is the current approach to detect a single date-format per column, then try to cast all dates to that?

Why can't we simply detect date vs datetime and the use `dateutil` to parse (since that supports a wide variety of formats)

maximlopin:
> I'm a bit confused by this to be honest. Namely, is the current approach to detect a single date-format per column, then try to cast all dates to that?
> 
> Why can't we simply detect date vs datetime and the use `dateutil` to parse (since that supports a wide variety of formats)

This issue: https://github.com/mindsdb/mindsdb_native/issues/471 was caused by dateutil being unable to parse dates
So now the approach is to check multifle formats one by one.
Another advantage of this approach is that we can expose date formats that we assumed in GUI (dateutil doesn't support that)

Most important pars of code to understand current approach:
1. Check if date: https://github.com/mindsdb/mindsdb_native/pull/495/files#diff-025d07f62339c33b46c5b5dddc027e7e8bd742999bcf4551154cf830fa10f086R168
2. If date, get its subtype: https://github.com/mindsdb/mindsdb_native/pull/495/files#diff-025d07f62339c33b46c5b5dddc027e7e8bd742999bcf4551154cf830fa10f086R87

George3d6:
Ok, makes sense, feel free to merge. But maybe make a note about supporting mixed date/datetime types in columns in the future.

PR #502, opened 2021-04-30
Optimize TypeDeductor for dates

maximlopin:


paxcema:
For what it's worth, I've made a few tests and this seems to be as fast as the pre-2.39 approach 👌

PR #507, opened 2021-05-20
Some time series fixes

paxcema:
## Why

To improve MindsDB accuracy and performance in time series tasks

## How

Changelog:
- Do not consider rows with incomplete historical context when calculating accuracy
- Changed data splitter criteria to have complete target info. for all rows (this is relevant when doing `t+n ; n>1` forecasts). A problem with the current behavior is that, depending on the granularity of the grouped series, null values can be passed as targets, yielding incorrect supervision in the predictor. This is fixed now.

### DIFF
--- a/mindsdb_native/libs/helpers/general_helpers.py
+++ b/mindsdb_native/libs/helpers/general_helpers.py
@@ -18,7 +18,6 @@ from sklearn.metrics import (
     r2_score
 )
 
-from lightwood.mixers import NnMixer
 from mindsdb_native.__about__ import __version__
 from mindsdb_native.config import CONFIG
 from mindsdb_native.libs.data_types.mindsdb_logger import log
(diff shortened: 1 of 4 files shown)


mindsdb_native/libs/phases/model_interface/lightwood_backend.py
```diff
@@ -503,13 +511,17 @@ def train(self):
 
                 if self.transaction.lmd['tss']['is_timeseries']:
                     validation_df = self.transaction.input_data.validation_df[self.transaction.input_data.validation_df['make_predictions'] == True]
+                    ts_window = self.transaction.lmd['tss'].get('window', 0)
+                else:
+                    ts_window = 0
```
George3d6: Why not make this None for non timesereis problems?

mindsdb_native/libs/helpers/general_helpers.py
```diff
@@ -203,6 +203,12 @@ def evaluate_regression_accuracy(
         backend,
         **kwargs
     ):
+    if kwargs.get('ts_window', None) is not None and backend.predictor and \
+            kwargs['ts_window'] <= 2*len(predictions[column]):
+        # with enough rows, truncate 1st 'window' outputs to account for incomplete historical context
+        predictions[column] = predictions[column][kwargs['ts_window']:]
+        true_values = true_values[kwargs['ts_window']:]
```
George3d6: This shouldn't trigger for none-timesereis problems, as it stands it will. Also, why check `backend.predictor` ?

George3d6 (changes requested):

mindsdb_native/libs/helpers/general_helpers.py
paxcema: Ok, `None` being the default avoids any non-TS task entering now. `backend.predictor` was a leftover, nice catch :+1:

